PR #2, opened 2016-03-20
labs instead of abs

stefan-balke:
When trying to build the lib on my mac, I got the error from the gcc
to use labs instead of abs.

I'm not strong in C, so does this make sense?

At least it was building the lib then...


erikd:
Thanks!

PR #3, opened 2016-07-02
Provide minimal CMakeLists

jcelerier:
Only tested in Win32 with Visual Studio 2015

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,14 @@
+cmake_minimum_required(VERSION 3.1)
+project(LibSamplerate LANGUAGES C)
+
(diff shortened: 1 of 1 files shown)


evpobr:
@jcelerier , it looks pretty nice. Some thoughts:
Win32 DLL will be broken (no exported functions). Add Win32/libsamplerate-0.def to source files under Win32 to fix it. 

Also you can set cmake_minimum_required to 3.0. There is nothing from version >= 3.0 in your script.

`/Win32` is `PRIVATE` include directory, no public headers inside.

Bobo1239:
This PR can be closed as #24 supersedes it.

PR #7, opened 2016-10-26
Fix configuration with MinGW under Windows

evpobr:
Running `./autogen.sh` with MinGW and Windows fails:

    $ ./autogen.sh
    checking for autoconf...
    checking for automake... yes
    checking for libtool... libtoolize
    I am going to run ./configure with no arguments - if you wish
    to pass any to it, please specify them on the ./autogen.sh command line.
    Generating configuration files for libsamplerate, please wait....
    aclocal -I M4
    ' is already registered with AC_CONFIG_FILES.
    ../../lib/autoconf/status.m4:288: AC_CONFIG_FILES is expanded from...
    configure.ac:312: the top level
    autom4te: /usr/bin/m4 failed with exit status: 1
    aclocal-1.15: error: echo failed with exit status: 1

This error happens because Git sets `CRLF` line endings under Windows in all files by default, but Autotools wants Linux-style `LF` line endings in its scripts.

This PR adds `.gitattributes` file and now Git will set proper `LF` line endings for Autotools scripts. Also some files were normalized and now have proper line endings (Windows makefiles etc).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,10 @@
+*.ac text eol=lf
+*.am text eol=lf
+*.m4 text eol=lf
+*.pc text eol=lf
+*.spec text eol=lf
+*.sh text eol=lf
+
+*.bat text eol=crlf
+*.def text eol=crlf
+*.msvc text eol=crlf


erikd:
The first commit is fine, but the second one couple easily be problematic. What happens when I use autotool's "make dist" command. Won't that just build a tarball with the line endings in my current directory (which will be unix line endings)?

erikd:
@evpobr I think you should look at the diffs before you submit a PR.

evpobr:
Yes, i'm working on it.

evpobr:
> Won't that just build a tarball with the line endings in my current directory (which will be unix line endings)?

I guess if you build one tarball under Windows with MinGW and other under Linux, tarballs will be different with or without this PR. Git will use `CRLF` for Win32 and `LF` for Linux automatically. Your `master` branch has `CRLF` line endings for me under Windows. And Autotools don't want it.

With my PR i want to have the same behaviour, BUT autotools stuff (see in `.gitattributes`) will have Linux line endings for everybody (Autotools happy), and Win32 files (*.def, *.bat, *.msvc) will always have `CRLF` (just to be consistent).

  Maybe you want Linux `LF` by default and set Win32 stuff `CRLF` explicitly? It is acceptable. This should fix MinGW problem 99,99% of Windows editors can handle `LF` (maybe only Notepad will be confused).

erikd:
There are bunch of Windows only files that should be CRLF always (Make.bat etc). For the rest, I don't mind local encodings in local checkouts as long as they get converted to unix line endings when they are committed.

evpobr:
This should work.

Windows stuff (`*.bat`, `*.def`, `*.msvc`) is `CRLF`, and also MinGW issue is fixed (Autotools files now have `LF`).

erikd:
I still don't understand why this PR includes anything other than the `.gitattibutes` file.

The files `Make.bat`, `Win32/Makefile.msvc` and `Win32/libsamplerate-0.def` already have the correct line endings in the repo. If I do a fresh clone on a Linux machine and then run the `file` utility on them I get:
```
> file Make.bat Win32/Makefile.msvc Win32/libsamplerate-0.def 
Make.bat:                  DOS batch file, ASCII text, with CRLF line terminators
Win32/Makefile.msvc:       makefile script, ASCII text, with CRLF line terminators
Win32/libsamplerate-0.def: ASCII text, with CRLF line terminators
```

evpobr:
> I still don't understand why this PR includes anything other than the .gitattibutes file.

I don't know.

This happens when i do this ([link](https://help.github.com/articles/dealing-with-line-endings/#refreshing-a-repository-after-changing-line-endings)).

erikd:
I'm not on my dev machine at the moment, but when I get in front of that, I'll merge the `.gitattibutes` patch and push it. You can then update and everything should be fine.

evpobr:
The work completed now.

erikd:
Appled. Thanks.

PR #8, opened 2016-10-30
More cmake work based on jcelerier's pull request.

mdsitton:
I tend to build cmake projects in a seperate build directory so files dont pollute the source tree, and it simplifies ignoring the cmake build files. This is mostly unrelated to the cmake changes, just explaining the .gitignore change.

WIN32 and _USRDLL i grabbed out of Makefile.msvc i'm not sure if they are actually required though.

This satisfies my needs, but if you want I can also look into getting the tests and examples building.


mdsitton:
I might also look into getting cmake to generate the config.h so other platforms work properly with cmake.

mdsitton:
Digging through the 2 workarounds in config,h 
```
#pragma warning(disable: 4305)
#define inline __inline
```
They don't seem to be needed anymore with Visual Studio 2015

mdsitton:
I based the cmake generated config file on the windows config, but didn't include anything that wasn't actually used by libsamplerate.

On windows for the FindSndfile an env variable should be set that points to the correct installations, though i'll need to set it up so it can handle both x86 and x86_64 versions.

erikd:
@mdsitton Thanks for do this. I would like to see a proper CMake based build system for libsamplerate, but I prefer not to merge it until its at least near being on par with the autotool based build system. That includes build examples and tests and running the tests.

mdsitton:
Alright, good to know, I'll start working on that next.

Bobo1239:
This PR can be closed as #24 supersedes it.

PR #12, opened 2017-04-26
Declare AUDIO_OUT once and in audio_out.h

coypoop:
Silences redefinition errors on some compilers
GCC 4.5.3 for some reason is more loud than other versions.
[Build log](http://nyftp.netbsd.org/pub/pkgsrc/packages/reports/2017Q1/NetBSD-6.0-i386/20170423.1108/libsamplerate-0.1.9/build.log)

### DIFF
--- a/examples/audio_out.c
+++ b/examples/audio_out.c
@@ -33,11 +33,6 @@
 #define MAKE_MAGIC(a,b,c,d,e,f,g,h)		\
 			((a) + ((b) << 1) + ((c) << 2) + ((d) << 3) + ((e) << 4) + ((f) << 5) + ((g) << 6) + ((h) << 7))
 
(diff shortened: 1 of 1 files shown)


examples/audio_out.h
```diff
@@ -6,7 +6,9 @@
 ** file at : https://github.com/erikd/libsamplerate/blob/master/COPYING
 */
 
-typedef	struct AUDIO_OUT_s AUDIO_OUT ;
+typedef	struct AUDIO_OUT_s
+{	int magic ;
+} AUDIO_OUT ;
```
erikd: With the definition of `AUDIO_OUT` removed from the file audio_out.c, this change should not be necessary.

Can you please try this and update the PR if it works?

coypoop:
That works!

PR #13, opened 2017-04-26
Remove dead assignment

coypoop:
I threw clang static analyzer at the code
float_increment is assigned again before it is used, making the initial assignment redundant.


erikd:
Will add a `travis.yml` file in the very near future.

PR #14, opened 2017-06-07
Autotools fixes

SoapGentoo:
* bzip2 tarballs
* Parallel testsuite (https://www.gnu.org/software/automake/manual/html_node/Parallel-Test-Harness.html)
* `--enable-werror` like libsndfile
* Do run make in `M4/` dir
* Don't start `./configure` right after `autogen.sh` (like libsndfile)
* Remove `--disable-gcc-pipe` and `--disable-gcc-opt` flags, respect user variables
* Try to use C99 mode for all compilers
* Massive Autoconf cleanup
* Don't use `$htmldocdir`, Autoconf provides `--htmldir`, which Automake also understands
* Use more portable POSIX pattern rules
* Fix GCC7 fallthrough errors
* Install `sndfile-resample` only if libsndfile is enabled and found


SoapGentoo:
@erikd I was thinking, don't you think just hard depending on C99 would simplify the build system a bit? It's 2017, if some obscure platform doesn't provide `lrintf()`, screw them. According to gnulib (https://www.gnu.org/software/gnulib/manual/html_node/lrintf.html), "FreeBSD 5.2.1, OpenBSD 3.8, Minix 3.1.8, AIX 5.1" do not provide an implementation of `lrintf()`. All of those OSes are absolutely vintage.

Issue #15, opened 2017-06-07
AUDIO_OUT related compilation failure


SoapGentoo:
When compiling on linux with ALSA, I get the following errors:
```
audio_out.c: In function ‘audio_play’:
audio_out.c:1014:16: error: dereferencing pointer to incomplete type ‘AUDIO_OUT {aka struct AUDIO_OUT_s}’
   if (audio_out->magic == ALSA_MAGIC)
                ^~
```

SoapGentoo:
@coypoop ideas?

SoapGentoo:
Fixed.

PR #16, opened 2017-06-10
Don't make AUDIO_OUT an opaque type

coypoop:
The linux ALSA code is using its members
Issue #15


SoapGentoo:
@coypoop See my fix in https://github.com/erikd/libsamplerate/pull/14/commits/65f62f99bae238bab408066daa2e798b2ee64854

coypoop:
That does seem like the intention, I didn't see it first, thanks!

PR #17, opened 2017-06-11
More fixes

SoapGentoo:


examples/audio_out.c
```diff
@@ -348,7 +348,13 @@ opensoundsys_open (int channels, int samplerate)
 		exit (1) ;
 		} ;
 
-	fmt = CPU_IS_BIG_ENDIAN ? AFMT_S16_BE : AFMT_S16_LE ;
+	fmt =
+#ifdef WORDS_BIGENDIAN
+	AFMT_S16_BE
+#else
+	AFMT_S16_LE
+#endif
```
erikd: Not keen on this change. Personally I consider `AC_C_BIGENDIAN` hugely flawed. In the past is has cost me a couple of hours debugging and I have not forgiven it.

The problem that macro is that if defines `WORDS_BIGENDIAN` and the suggested usage is as you have used it here. The problem is that if you use it like this and forget to include `<config.h>` it will *sliently* do the wrong thing on big endian platforms which are far less tested anyway.

SoapGentoo: @erikd I understand the issue, and yes, that is a misdesign on the part of Autoconf. I have two suggestions:

1. I can test it on our Gentoo big-endian machines (alpha, sparc, ia64) and check that indeed the right `#ifdef` clause is invoked, _or_
2. instead of completely bringing back the old macro, have a look at the macro declaration:

	`AC_C_BIGENDIAN([action-if-true], [action-if-false], [action-if-unknown])`

   we add your old `BIG_`/`LITTLE_` endian macros in the `action-if-true` and `action-if-false` blocks. This way we get rid of the old code, and have the explicitly-required defined macro as you have argued for. Let me know which variant you would like me to implement.

I personally prefer 1. as its more idiomatic (and am willing to test it), but I see your point.

erikd: 1. is only a solution right now it is not a long term solution.

I can however live with solution 2.

SoapGentoo: @erikd please have another look, fixed now using approach 2.

erikd:
:+1:

PR #22, opened 2017-07-24
Fix #18

### LINKED ISSUE #18: Const-correctness?
Is this OK:

    SRC_DATA data{};
    data.data_in=const_cast<float*>(buffer_in);

That is, can I be sure that the library does not touch buffer_in? If so add `const`, or at least annotate that in the documentation.

milasudril:
* Update api_full.html
 * Update api_misc.html
 * Update api_simple.html

### DIFF
--- a/doc/api_full.html
+++ b/doc/api_full.html
@@ -111,7 +111,8 @@ function has the following fields:
 </P>
 <PRE>
       typedef struct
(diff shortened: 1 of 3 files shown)


erikd:
You still need some more git/github/PR practice, but I'll take this as is :).

Thanks.

PR #23, opened 2017-10-04
Add ability to clone a SRC_STATE* handle

hoyes:
I had a use case where I wished to copy a handle along with all its filter state, so that resampling could be resumed elsewhere without discontinuities. Pulling in case this would be relevant to others.


erikd:
This looks reasonably sane, but I would love to see a test for it added to the test suite.

hoyes:
Test, documentation and some missing export definitions added above.
I've attempted to make the necessary modifications to the files in Win32/* by inspection but haven't been able to test the changes there.

erikd:
Manually merged. Thanks!

PR #24, opened 2017-11-26
Add CMake build system (based on previous work from @mdsitton and @jcelerier) and CI

Bobo1239:
Hi @erikd,
thanks for creating this library. I wanted to create some Rust bindings and to simplify cross-platform compilation I continued the previous CMake attempts from @mdsitton and @jcelerier. I've also tried to setup some CI (Travis for Linux; Appveyor for Windows). Admittedly I've not really done much C development till now so please feel free to point out any problems with this PR.

https://travis-ci.org/Bobo1239/libsamplerate
https://ci.appveyor.com/project/Bobo1239/libsamplerate


erikd:
This is a pretty large PR. Probably won't get time to have a proper look at it until the weekend.

jcelerier:
hey, sorry, I had entirely forgot my PR. this looks pretty nice and complete ! But are you sure you want to support CMake versions that didn't even have PUBLIC / PRIVATE ? Are there still widely-used systems with such old version (that couldn't just install CMake's statically-built latest version) ?

Bobo1239:
@erikd No worries, take your time. :)
@jcelerier Tbh I was just trying to get everything [here](https://travis-ci.org/Bobo1239/libsamplerate-rs/builds/308217378) to become green ([as they are now](https://travis-ci.org/Bobo1239/libsamplerate-rs/builds/308241215)) and some of those platforms (they are [Docker images](https://github.com/japaric/cross/tree/master/docker)) have CMake 2.8. I figured we aren't really using any new CMake features so it wouldn't hurt to support an older verson. But as I said in my opening post I don't have an overview over this ecosystem so you guys have to make the call whether it's worth to support old CMake versions.

erikd:
[libsndfile](https://github.com/erikd/libsndfile/) builds on travis with CMake min version of 3.1.3. Maybe figure out what that is doing.

Bobo1239:
Those builds aren't normal Travis instances; they are cross compiling and testing on many different Docker images (some of them using qemu internally?) and aren't maintained by me. (They are from [cross](https://github.com/japaric/cross) which is ""Zero setup" cross compilation and "cross testing" of Rust crates"). If you guys don't think it's worth supporting such an old CMake version I have no problems reverting the last few changes. I would then just drop those few Travis instances or try to upgrade CMake in those Docker images.

erikd:
Yeah, I don't think its worth supporting any CMake version older than the ones libsndfile uses.

jcelerier:
that's interesting: these docker images have rust which is very recent, but don't have five-years old versions of CMake :p are they hosted somewhere on github ? it could be worth it to send PRs to them so that everyone doing such a work can at least leverage recent versions.

Bobo1239:
Reverted the last changes.
The docker images are [here](https://github.com/japaric/cross/tree/master/docker). They are using different Ubuntu base images (ranging from 12.04 to 16.04) so some of the older ones don't have CMake 3. I'll try to naively upgrade some of the Ubuntu versions but I'm expecting some issues as there's probably a reason that they are using different versions.

tresf:
@Bobo1239 do you offer any build artifacts for `libsamplerate`?  If not, would the project entertain an automatic upload (e.g. github releases, etc?) to complement the cmake support?

I'm most interested in the win32 artifacts for multi-platform.  The other OSs (unix/linux/macos) have methods to fetch `libsamplerate`.  CMake drastically simplifies the MSVC++ setup and I'd like to add the 32-bit and 64-bit artifacts into the `scoop.sh` package manager.

CMakeLists.txt
```diff
@@ -0,0 +1,119 @@
...
+check_type_size(long SIZEOF_LONG)
+
+# This will set CPU_CLIPS_NEGATIVE and CPU_CLIPS_POSITIVE
+clip_mode()
+
+find_package(ALSA)
+set(HAVE_ALSA ${ALSA_FOUND})
+
+find_package(Sndfile)
+set(HAVE_SNDFILE ${SNDFILE_FOUND})
```
tresf: @Bobo1239 cmake does not seem to provide `FindSndfile.cmake` or `FindFFTW.cmake` on some systems which means if they're in a non-standard location, the linking and including won't work as it would in a vanilla POSIX environment.

Would you consider adding an additional `pkg-config` dependency on this?  Then you could do the following:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 394de75..1b85d17 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,7 @@ include(CheckIncludeFiles)
 include(CheckSymbolExists)
 include(CheckTypeSize)
 include(ClipMode)
+include(FindPkgConfig)

 set(SAMPLERATE_SRC
        ${PROJECT_SOURCE_DIR}/src/samplerate.c
@@ -53,12 +54,24 @@ clip_mode()

 find_package(ALSA)
 set(HAVE_ALSA ${ALSA_FOUND})
+if(ALSA_FOUND)
+       include_directories("${ALSA_INCLUDE_DIRS}")
+       link_directories("${ALSA_LIBRARY_DIRS}")
+endif()

-find_package(Sndfile)
+pkg_check_modules(Sndfile sndfile>=1.0.0)
 set(HAVE_SNDFILE ${SNDFILE_FOUND})
+if(SNDFILE_FOUND)
+       include_directories("${SNDFILE_INCLUDE_DIRS}")
+       link_directories("${SNDFILE_LIBRARY_DIRS}")
+endif()

-find_package(FFTW)
+pkg_check_modules(FFTW3 fftw3>=3.0.0)
 set(HAVE_FFTW3 ${FFTW_FOUND})
+if(FFTW3_FOUND)
+       include_directories("${FFTW3_INCLUDE_DIRS}")
+       link_directories("${FFTW3_LIBRARY_DIRS}")
+endif()

 configure_file(${PROJECT_SOURCE_DIR}/config.h.in ${CMAKE_CURRENT_BINARY_DIR}/config.h)
```

Bobo1239: `FindSndfile.cmake` and `FindFFTW.cmake` are the `cmake` directory which should be loaded by `set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR}/cmake)` at the top. (`ClipMode.cmake` is also affected by this.)

Regarding pkg-config: What do @erikd and @jcelerier think? Should I switch from the two cmake modules to pkg-config? I really don't know what is better supported across the board.

Bobo1239:
@tresf I'll try to look into AppVeyor build artifacts this weekend.

CMakeLists.txt
tresf: >  I really don't know what is better supported across the board.

To give a bit more background information on this:

Ubuntu 14.04, default for Travis-CI:

```bash
locate -i FindALSA.cmake
(found) /usr/share/cmake-3.5/Modules/FindALSA.cmake
```

```bash
locate -i FindFFTW
(not found)
```

```
locate -i FindSndfile
(not found)
```

What this means is if a project requires any of the variables expected to be available for includes or libraries, they won't be available.  Msys2 re-uses the POSIX-style paths so it hides this bug as the default `include` and `lib` paths work out of the box.

tresf: @Bobo1239 On second thought, perhaps this is a problem with my environment.  [`CMAKE_PREFIX_PATH`](https://cmake.org/cmake/help/v3.0/variable/CMAKE_PREFIX_PATH.html) clearly states that it searches the same paths as `pkg-config`.  Here's an excerpt from the website:

> *[CMAKE_PREFIX_PATH] contains the “base” directories, the `FIND_XXX()` commands append appropriate subdirectories to the base directories. So `FIND_PROGRAM()` adds `/bin` to each of the directories in the path, `FIND_LIBRARY()` appends `/lib` to each of the directories, and `FIND_PATH()` and `FIND_FILE()` append `/include` .*

I will investigate this some more on my end as to why the paths are found, but the includes and libs are not.

Bobo1239: I think you may have misunderstood me. FindFFTW and FindSndFile are distributed with the source (in the `cmake` directory) in the repo and aren't required from the environment.

cmake/FindFFTW.cmake
```diff
@@ -0,0 +1,25 @@
+# From: https://github.com/wjakob/layerlab/blob/master/cmake/FindFFTW.cmake
+# FIXME: License concerns?
```
tresf: The license is `BSD 2-clause "Simplified" License`.  This and the BSD 3-clause are GPL compatible and very common for cmake scripts.

cmake/FindFFTW.cmake
```diff
@@ -0,0 +1,25 @@
...
+#  FFTW_INCLUDES    - where to find fftw3.h
+#  FFTW_LIBRARIES   - List of libraries when using FFTW.
+#  FFTW_FOUND       - True if FFTW found.
+
+if (FFTW_INCLUDES)
+  # Already in cache, be silent
+  set (FFTW_FIND_QUIETLY TRUE)
+endif (FFTW_INCLUDES)
+
+find_path (FFTW_INCLUDES fftw3.h)
```
tresf: Ok... I think I found the problem.  I would expect this to be `FFTW_INCLUDE_DIR` and `FFTW_LIBRARY_DIRS`.  See `FindSndfile.cmake` as an example.  This is mostly likely why `pkg-config` worked and `find_package` failed.

CMakeLists.txt
tresf: @Bobo1239 you are correct, I missed that they were distributed with the project. I think I found the culprit: https://github.com/erikd/libsamplerate/pull/24/files#r157309049.  Your original technique should be just fine but I recommend making the variables consistently named.  Also, `include_directories(...)` and `link_directories(...)` will still be required for some environments.

cmake/FindFFTW.cmake
erikd: I'd be fine with either a BSD or BSD 2-clause license for the CMake files.

CMakeLists.txt
tresf: I looked at some more `FindXXX.cmake` modules and they all appear to follow slightly different conventions.  Here's what works out of the box using this PR without adding `pkg-config`:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 394de75..a2aad7f 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,11 +53,23 @@ clip_mode()

 find_package(ALSA)
 set(HAVE_ALSA ${ALSA_FOUND})
+if(ALSA_FOUND)
+  include_directories("${ALSA_INCLUDE_DIR}")
+  link_directories("${ALSA_LIBRARY}/../")
+endif()

 find_package(Sndfile)
 set(HAVE_SNDFILE ${SNDFILE_FOUND})
+if(SNDFILE_FOUND)
+  include_directories("${SNDFILE_INCLUDE_DIR}")
+  link_directories("${SNDFILE_LIBRARY}/../")
+endif()

 find_package(FFTW)
+if(FFTW_FOUND)
+  include_directories("${FFTW_INCLUDES}")
+  link_directories("${FFTW_LIBRARIES}/../")
+endif()
 set(HAVE_FFTW3 ${FFTW_FOUND})

 configure_file(${PROJECT_SOURCE_DIR}/config.h.in ${CMAKE_CURRENT_BINARY_DIR}/config.h)
```

tresf:
@Bobo1239 good call on `target_link_libraries`.  a1ba332 now fixes building on non-posix environments.  Thanks!

Bobo1239:
@tresf Would [this](https://ci.appveyor.com/project/Bobo1239/libsamplerate/build/job/wntvku14cmyvo3ns/artifacts) be enough for your use case?

tresf:
@Bobo1239 will be needing the header file and the pkg-config (`.pc`) file.  [`libsndfile`](http://www.mega-nerd.com/libsndfile/files/libsndfile-1.0.28-w64.zip) is a good example and here's how I've written the release [into the scoop package manager](https://github.com/tresf/scoop-extras/blob/7fbf5b4fd555689805132541605f00851f042272/libsndfile.json).

Also, long-term, if there are no objections, it would be nice to have the file hosted at an unauthenticated, tagged location such as the github release area, if @erikd would entertain this.  [AppVeyor is able to do this](https://www.appveyor.com/docs/deployment/github/).

Bobo1239:
@tresf I guess it would make sense to setup the install target in CMake so we get that kind of structure for all platforms (also we're currently not generating a `.pc` file at all I think?)
I don't know when I'll have the time to try that out. If you need it rather sooner than later we can maybe try to land this PR first and let somebody else do that?

tresf:
@Bobo1239 AFAIK `.pc` file should look like this:

```cmake
prefix=
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: samplerate
Description: An audio Sample Rate Conversion library
Requires: 
Version: 0.1.8
Libs: -L${libdir} -lsamplerate
Libs.private: Ext/libsndfile.a
Cflags: -I${includedir} 
```

CMake can configure this file like so (sample, I haven't tested it for syntax errors)
```cmake
SET(prefix "")
SET(exe_prefix "\\${prefix}")
SET(libdir "\\${exec_prefix}/lib")
SET(includedir "\\${prefix}/include")
SET(VERSION "${PROJECT_VERSION}")
SET(LIBS "Ext/libsndfile.a")
CONFIGURE_FILE(samplerate.pc.in samplerate.pc @ONLY)
```

tresf:
> @tresf I guess it would make sense to setup the install target in CMake so we get that kind of structure for all platforms 

I agree, although I can rearrange the files in the packaging utility if needed, that's [what we have to do with fftw's](https://github.com/tresf/scoop-extras/blob/678f5699403276e590664cb8c39ba983182b23a9/fftw.json#L33) for example.  Getting the output files is enough to start offering this package to win32 developers.

Bobo1239:
https://ci.appveyor.com/project/Bobo1239/libsamplerate/build/job/psucvyvat0tv6vye/artifacts Does this work?

tresf:
@Bobo1239 Yes, it works quite well actually. :)

```diff
  -- Checking for module 'samplerate>=0.1.8'
+ --   Found samplerate, version 0.1.9
```

A working scoop.sh bucket is [here](https://gist.github.com/tresf/a5dc944dc9c5d437261b1e38171b1ac1#file-libsamplerate-json-L10).

Bobo1239:
@erikd Is there anything I should do here?

tests/util.c
```diff
@@ -176,12 +176,10 @@ get_cpu_name (void)
...
 	is_pipe = 1 ;
 #else
-	file = NULL ;
+	return name;
 #endif
 
-	if (file == NULL)
-		return name ;
-
+#if defined (__linux__) || defined (__APPLE__) || defined (__FreeBSD__)
```
erikd: Something's not right there. Why has that `#if` been removed?

erikd:
Would you be fix the `#if` issue and then squash that down to a single commit crediting the other two contributors and giving a good overview commit message?

tests/util.c
Bobo1239: The problem here is the `pclose` call afterwards which fails to link on Windows. An alternative would be to replace `is_pipe` with `#define IS_PIPE` (+ `#undef IS_PIPE` at the end) and
```
if (is_pipe)
	pclose (file) ;
else
	fclose (file) ;
```
with
```
#ifdef IS_PIPE
	pclose (file) ;
#else
	fclose (file) ;
#endif
```
Do you want me to do that?

erikd: What problem was this change trying to address?

Bobo1239: Native Windows (MSVC) doesn't provide `pclose` (though apparently it has [`_pclose`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/pclose)) so it must be removed by the preprocessor.

erikd: Ok,  how about you just sqash this PR down to a single commit and I'll sort it out,

Bobo1239:
Sorry for the delay. Finally got around to do this. (Was busy with university...)
I'm not sure if this qualifies as a "good overview commit message" so tell if there's anything you wish me to add. Besides a rebase onto master I haven't done any changes to the code.
In case you're merging this: don't forget to enable AppVeyor for the repo (and PRs?).

erikd:
Thank you all for your contribution!

PR #25, opened 2017-11-29
make it build using c89 compilers

sezero:
These changes make libsamplerate to build using c89 compilers (e.g. Open Watcom).


erikd:
Sorry, its the year 2017. I do not support C compilers that are not C99 compliant.

sezero:
> its the year 2017.

I'm aware,

> I do not support C compilers that are not C99 compliant.

OK, as you wish.

sezero:
[]sorry that I saw this part later

> Is that Watcom compiler?

Yes, Open Watcom.

> On which OS?

I am running on Linux, targeting OS/2 and Windows.

PR #26, opened 2017-11-29
add Watcom versions of lrint and lrintf asm inlines

sezero:
This changeset adds lrint() and lrintf() 386 asm inlines for Watcom compiler.


erikd:
Thanks!

PR #27, opened 2018-01-05
drop comma at end of enumerator

janstary:
Some older compilers might still choke on this with -pedantic.
And it saves a patchfile downstream.

### DIFF
--- a/src/samplerate.h
+++ b/src/samplerate.h
@@ -166,7 +166,7 @@ enum
 	SRC_SINC_MEDIUM_QUALITY		= 1,
 	SRC_SINC_FASTEST			= 2,
 	SRC_ZERO_ORDER_HOLD			= 3,
-	SRC_LINEAR					= 4,
+	SRC_LINEAR					= 4
 } ;
 
 /*


erikd:
What compiler is it that doesn't handle this? Does that compiler not have a C99 compilation mode?

janstary:
Myself, I don't use any compiler the could not handle this.
The MacPorts project currently patches that away (not sure if it's needed).

I maintain the port, and would like to simplify it by getting rid of the patch.
I believe it's a tiny syntax change that makes such considerations obsolete.

janstary:
https://trac.macports.org/ticket/28324

erikd:
That patch is from 7 years ago. Are you sure its still needed? If it is, its almost certainly because the right flags aren't being passed to the compiler.

janstary:
I am not sure it's still needed; in fact, I believe it's probably not.
But still, a dependent port can compile with -Werror -pedantic and choke
while compiling its libsamplerate dependency, for no good reason.

I honestly believe that removing that one char would be
the simplest way to get rid of this possible-non-problem.

erikd:
I think people using `-pedantic` without specifying a C standard version is a much bigger problem than this comma. I think its worth keeping this comma just to teach that lesson.

erikd:
If Macports is using `-pedandtic` by default and then adding a patch to remove the comma, then they should fix that. Will therefore not be applying this patch.

Closing.

PR #29, opened 2018-01-05
BSD lincensed now

janstary:
Libsamplerate is BSD licensed since 2016.
(So we should get rid of the GNU auto* cruft :-)


erikd:
Applied. Thanks!

Issue #33, opened 2018-01-14
src_process rejects data_in = NULL even when input_frames = 0


enzo1982:
Hi Erik,

I found that `src_process` throws an error when called with `data_in = NULL` and `input_frames = 0`.

I was calling it like that with `end_of_input` set to finish a resampling process. I think it's a bit unintuitive that this doesn't work and the documentation also does not indicate that it might be forbidden.

A work-around is to set `data_in` to a dummy value like `(float *) 1` in the special case of finishing the resampling process with no more frames to resample.

So my suggestion is to accept `data_in = NULL` when `input_frames = 0` or, alternatively, mention in the API docs that `data_in` must not be `NULL` even when passing no input frames. I think just accepting `NULL` in this case would be the better solution, though.

Here is an example program that illustrates the issue: [finish-test.c.gz](https://github.com/erikd/libsamplerate/files/1629289/finish-test.c.gz)

Thanks,
Robert

PR #34, opened 2018-01-14
Accept data_in = NULL when passing no input frames

enzo1982:
Proposed change to fix issue #33.


erikd:
Merged manually. Thanks!

PR #35, opened 2018-01-27
Move `sndfile-resample` to sndfile-tools package

SoapGentoo:
* Fixes https://bugs.gentoo.org/645258


SoapGentoo:
@erikd anything else to do here?

erikd:
There's still libsndfile stuff in the `configure.ac`, `Win32/Makefile.mingw.in` and `INSTALL`.

PR #38, opened 2018-04-20
Disable examples when alsa/libsndfile are not present

SoapGentoo:
@erikd This also solves problems for users when ALSA and/or libsndfile aren't installed. Instead of erroring out, the targets are skipped gracefully.


SoapGentoo:
@erikd done

PR #47, opened 2018-10-04
Adds CMake option to enable/disable tests, default enabled when building standalone

robiwano:


erikd:
This does not pass CI. If you fix it, I will be merged.

CMakeLists.txt
```diff
@@ -1,6 +1,9 @@
 cmake_minimum_required(VERSION 3.1)
 project(libsamplerate VERSION 0.1.9 LANGUAGES C)
 
+option(LIBSAMPLERATE_TESTS ON "Enable to generate test targets")
```
qno: setting this option is wrong - it must be changed to 

`option(LIBSAMPLERATE_TESTS "Enable to generate test targets" ON)`

see https://cmake.org/cmake/help/latest/command/option.html

CMakeLists.txt
```diff
@@ -1,6 +1,9 @@
...
+option(LIBSAMPLERATE_INSTALL ON "Enable to add install directives")
```
qno: dito

CMakeLists.txt
robiwano: Indeed, thanks!

erikd:
@robiwano Would you be able to squash these two commits into a single one?

erikd:
Applied. Thanks!

PR #48, opened 2018-10-08
Prevent a couple div by zero scenarios.

hnand:


erikd:
This is un-necessary. The `src_ratio` is already limited in such a way that `0.0` is not possible by the use of https://github.com/erikd/libsamplerate/blob/master/src/common.h#L157-L160

hnand:
Ahh, I see. Sounds good. I will debug this further. Thank you for the explanation.

PR #50, opened 2018-11-12
add missing src_get_channels export to windows def file

sezero:


erikd:
This was merged as part of #51.

Thanks!

Issue #28, opened 2018-01-05
examples/audio_out.c fails on newer MacOS


janstary:
`examples/audio_out.c` fails to build on MacOS 10.13.2,
because it includes `Carbon.h` which does not exist.
Apparently, the MacOS audio subsystem has changed again
(to proactively leverage the synergy of the frameworks).

janstary:
Currently, the MacOS port just does not build the examples.
https://github.com/macports/macports-ports/blob/master/audio/libsamplerate/files/patch_examples_Makefile.in.diff

erikd:
I am well aware  of this. I don't use Mac OS X so someone else needs to write this.

jmroot:
First problem by inspection is AudioHardwareGetProperty, which went away after 10.6. Migration guide here: https://developer.apple.com/library/content/technotes/tn2223/_index.html

conradjones:
pull request https://github.com/erikd/libsamplerate/pull/53

erikd:
Fix has been merged. Closing this. Thanks!

PR #53, opened 2019-02-07
fix macos compile errors - modernize audio api

conradjones:
reference issue #28

code now compiles on modern macOS.

I found forums posts stating the old API was deprecated dated 2008 so I don't believe supporting the old and new is necessary, happy to be corrected though.


erikd:
This looks great. I would just ask you to squash these two commits into a single commit.

conradjones:
done

PR #54, opened 2019-03-05
Check for and suppress int shortening warnings

danra:
- Came across these warnings when compiling on Xcode with this warnings turned on. 
- Reproduced in command-line build by adding the appropriate warning flag to Makefile.ac.
- Suppressed the warnings by adding explicit casts - all of them seemed to me to be benign.


erikd:
Thanks!

PR #62, opened 2019-05-29
drop libsndfile and alsa

janstary:
* remove the examples, which use sndfile
* drop the libsndfile dependency
* drop alsa, as nothing uses it now
* remove all mentions of sndfile, alsa and examples/ from the build
* remove all mentions of sndfile, alsa and examples/ from the doc
* do not touch the win directories (leaving it to win users )
* remove dead links from the doc while here

Tested on MacOS 10.14.5


janstary:
Also tested on OpenBSD 6.5/amd64: builds and installs fine,
`make test` says `ok` or `pass` for all tests.

janstary:
Also tested on FreeBSD 12.0/amd64: builds and installs fine,
`make test` says `ok` or `pass` for all tests.

erikd:
Why? Those examples were added because when this code was first released, people we asking for examples on how to use it.

janstary:
```
On Oct 10 20:23:49, erikd@mega-nerd.com wrote:
> Jan Stary wrote:
> > Currently, the sndfile-resample utility that comes with SRC
> > does not have a manpage. See below for one that extends usage();
> 
> Actually, since that program is moving to the sndfile-tools 
> project (so that libsamplerate can drop the libsndfile dependency)
```

And everything in examples/ depends on libsndfile.

erikd:
Yes, but `sndfile-resample` grew from being an example program to something that was useful in its own right. Hence it moved the the `sndfile-tools` repo/project. The other example programs in the `examples/` directory are no useful in their own right, but are good examples and I think worth keeping (regardless of what I may have written before).

janstary:
>> (regardless of what I may have written before).

Well if the libsndfile dependency is staying,
the whole premise of this change is wrong.

erikd:
> Well if the libsndfile dependency is staying, the whole premise of this change is wrong.

Yes, I am afraid that is correct.

janstary:
See instead https://github.com/erikd/libsamplerate/pull/63

Issue #68, opened 2019-07-23
Require C99 compiler also in CMake


Flamefire:
The autoconf already requires a C99 compiler. I suggest to do the same for CMake which simplifies the code, config and removes the whole float_convert header.

Flamefire:
Closed by #81

PR #69, opened 2019-07-23
CMake improvements

### LINKED ISSUE #46: Make CMake tests and install directives optional
When using libsamplerate via CMake and as a submodule, all targets related to tests/examples clog up generated workspace. It should be make optional f.i. with:

`option(LIBSAMPLERATE_TESTS OFF "Enable to generate test targets)
`

Same applies to `install(...)` directives.

Flamefire:
I looked over the CMake files and implemented a couple of enhancements and fixes.

- Modularize into folders (test, examples, include, src) and add correct defaults (closes #46)
- Make config.h smaller and divide into 3 configs (examples, test, lib)
  - Allows to see what is actually required or define own config options
- Move public include into include folder

This should now allow to use this project as a submodule from CMake with `add_subdirectory` allowing broader use without pulling in unnecessary dependencies or checks (FFTW, sndfile...)


tests/throughput_test.c
```diff
@@ -74,7 +74,7 @@ throughput_test (int converter, long best_throughput)
 		exit (1) ;
 		} ;
 
-	throughput = lrint (floor (total_frames / duration)) ;
+	throughput = (long) floor (total_frames / duration) ;
```
erikd: You are chaning the code code here.  The oepration of `lrint` is *different* from the casting to `long`.

tests/util.h
```diff
@@ -29,7 +29,7 @@ double calculate_snr (float *data, int len, int expected_peaks) ;
 
 const char * get_cpu_name (void) ;
 
-#if OS_IS_WIN32
+#ifdef WIN32
```
erikd: Using `#ifdef` is less safe than using `#if`. If the mechanism that defines a variable used like this goes wrong, it can fail silently.

The variable `OS_IS_WIN32` is defined to be either `0` or `1`. If the mechanism to define it goes wrong, you get a compile error.

tests/multichan_throughput_test.c
```diff
@@ -74,7 +74,7 @@ throughput_test (int converter, int channels, long best_throughput)
 		exit (1) ;
 		} ;
 
-	throughput = lrint (floor (total_frames / duration)) ;
+	throughput = (long) floor (total_frames / duration) ;
```
erikd: Not the same.

src/float_cast.h
```diff
@@ -36,40 +36,24 @@
...
 #include "config.h"
+#include <math.h>
 
 /*
 **	The presence of the required functions are detected during the configure
 **	process and the values HAVE_LRINT and HAVE_LRINTF are set accordingly in
 **	the config.h file.
 */
 
-#define		HAVE_LRINT_REPLACEMENT	0
```
erikd: No, this is not going to be removed.

configure.ac
```diff
@@ -277,8 +272,6 @@ AS_CASE([${host_os}],
 dnl ====================================================================================
 dnl  Now use the information from the checking stage.
 
-AC_DEFINE_UNQUOTED([OS_IS_WIN32], [${os_is_win32}], [Set to 1 if compiling for Win32])
```
erikd: This is significantly better than what you want to replace it with.

Makefile.am
```diff
@@ -19,7 +19,7 @@ test:
 ########
 
 lib_LTLIBRARIES = src/libsamplerate.la
-include_HEADERS = src/samplerate.h
+include_HEADERS = include/samplerate.h
```
erikd: What???? Its fine (desirable even) to keep the header file in `src/` and during installation put it in `/usr/include`.

Makefile.am
```diff
@@ -157,10 +157,6 @@ tests_multichan_throughput_test_SOURCES = tests/multichan_throughput_test.c test
 tests_multichan_throughput_test_CFLAGS = $(FFTW3_CFLAGS)
 tests_multichan_throughput_test_LDADD = src/libsamplerate.la $(FFTW3_LIBS)
 
-if HAVE_LIBSNDFILE
-check_PROGRAMS += tests/src-evaluate
```
erikd: Err, no.

Flamefire:
> This is a whole bunch of unrelated changes. You should probably PR some of these separately. 

Sure. Which ones would you have separately? Most of the things is very related so I just put them in separate commits. What I can think of is the math/clip fixes and the src-evaluate fix although the cmake part depends on this one and I don't really like dependent MRs. So just tell me what you'd like in a separate MR.

> You should also understand why things are the way they are before you decide to change them.

I do. I'm proficient in C++ & CMake but might miss some parts on C and autotools. So I got good reasons to change things but the solution might not be optimal. Let's talk about the individual changes to make this great project even better.

tests/throughput_test.c
Flamefire: Can you tell me why? Reason for the change is I wanted to avoid the (private) header float_cast.h. `lrint` rounds according to current rounding mode, then converts to long. Rounding is done here by `floor` so only remaining thing is the conversion. I'm aware that `lrint` might be faster, but a) this is a test and b) it is executed rarely (once? once per individual test? Not in a hot loop anyway)

tests/util.h
Flamefire: I agree here and like that error too (used in our projects too). However in this case the complexity in configuring this could be easily replaced by existing macros which are defined for exactly this purpose. I could/should enhance this to `if defined(_WIN32) || defined(__WIN32__) || defined(WIN32)` to catch more compilers (I think `_WIN32` is the most correct one, the rest is compatibility)

As this is in tests though I don't really care which way it is done. Your call

tests/multichan_throughput_test.c
Flamefire: See above reasoning https://github.com/erikd/libsamplerate/pull/69#discussion_r306711585

src/float_cast.h
Flamefire: Why? It is never used and not a standard macro. this header is also deprecated (I'd rather remove the whole header, see #68) so why keep it? What could it be used for?

Makefile.am
Flamefire: I disagree here. Could you tell me your reasoning why it would be "even desirable" to keep the public header in `src/`?

What proved advantageous was to separate out public and private headers already in the source tree. Public headers (ones installed) go into `include` private ones into `src`. This has 2 advantages:
- Tests and examples can be build against the source tree. See https://github.com/erikd/libsamplerate/pull/69/files#diff-af3b638bc2a3e6c650974192a53c7291R70 The binaries linked against libsamplerate don't get to include the private headers. This avoids accidentally using a private there which would break on "real installs" and hence is an additional safe guard.
- Consumers can add this git repo as a submodule and simply use it with `add_subdirectory` (motivation for this PR) They get the same includes (only public ones) as if using the installed library. This avoids having to install this in a separate step (and with a possibly mismatching compiler setting) which is a big bonus especially for Windows where dependencies are just a nightmare.

Makefile.am
Flamefire: Why? **ALL** your tests/examples are written in a way, that they compile to stubs if the required libs are not found. So why have a special case for this one? It works this way (proven by CI)

The current approach fails because of https://github.com/erikd/libsamplerate/pull/69/files/8f19f25358ead6fd7bb43f7099e228265b72ae3e..1c7b71e07e5610ea0be9504c485f687e9aec03d4#diff-c949f93d03f44a4217d7a138f9e2e54aR82

The program is unconditionally required and if libsndfile isn't found it breaks the built as no sources are added to that binary. The alternative to this change is to remove it from the linked `check_PROGRAMS` but after seeing the `#if HAVE_LIBSNDFILE` in the source, the indented way looks rather like I have done it here.

tests/throughput_test.c
Flamefire: Alternative is to require C99 also in CMake: #68

Flamefire:
I created smaller PRs which should be merged first to make this one small:
- [ ] #73
- [ ] #74
- [ ] #75
- [ ] #76

tests/util.h
Flamefire: I updated this to use the old define. As this is test code it doesn't matter if configure is required.

configure.ac
Flamefire: Reverted

Makefile.am
Flamefire: Pulled into separate PR #73 to fix bug described here

Flamefire:
I updated this PR to remove changes that were moved into other PRs and addressed comments. Anything else that should be factored into its own PR? Would be great to get this in, so projects using this are not impacted by dependencies for tests (e.g. searching for FFTW, libsndfile...)

evpobr:
@Flamefire

> Move public include into include folder

I like it.

> Make config.h smaller and divide into 3 configs

Not sure. It is common practice to have one `config.h`. This is a support file, don't spend too much time on it. But that `src/src_config.h` should die. Usually it is simply used to include a `config.h` wrapped in a `#ifdef HAVE_CONFIG_H`. 

> Modularize into folders (test, examples, include, src)

Not sure. It seems to me that these are not such large projects to divide them into subdirectories. We have moved away from recursive makefiles and CMake files in libsndfile, to avoid a bunch of files here and there, it would be strange to have a project with subdirectories here.

Flamefire:
> Not sure. It is common practice to have one config.h. This is a support file, don't spend too much time on it. But that src/src_config.h should die. Usually it is simply used to include a config.h wrapped in a #ifdef HAVE_CONFIG_H.

Sure but that config.h usually contains only macros required to make sure only features actually contained in the binary so consumers don't use stuff that isn't there. In this case (IIRC) there is more. Especially mixing in stuff that is only required for the tests is IMO reason enough to split it. Or to remove the test-only stuff and use plain preprocessor definitions on command line
The src_config.h was added by someone else who doesn't want a configure step IIRC. Fine by me.

> Not sure. It seems to me that these are not such large projects to divide them into subdirectories. 

It isn't that much about size but about organization. E.g. some dependencies are only required for tests or examples. With the 1-folder approach the boundaries are hard to define, with subdirs it is easy. (Example: With CMake I see it searching for FFTW and that might confuse new people into thinking FFTW is required for this library to work or provide some benefit)
And for users having an "examples" folder is also nice, so they can get a quick look how to use it. The tests are often not suitable for it. But tests and examples could be combined into 1 folder if wanted and files are properly named.

evpobr:
Ok, let's start from `include` PR. Add then create CMake subdirectories PR. Perhaps it is more convenient than I think now.

evpobr:
> Sure but that config.h usually contains only macros required to make sure only features actually contained in the binary so consumers don't use stuff that isn't there. In this case (IIRC) there is more. Especially mixing in stuff that is only required for the tests is IMO reason enough to split it. Or to remove the test-only stuff and use plain preprocessor definitions on command line

I suggest another option - why not move the checks required for the library and tests to their subprojects, and call `confgure_file`() after all `add_subdirectory`()?

Flamefire:
Given that config.h is not installed this is certainly an option. The downside I see is that this way it can't be tested whether the main header is "clean", i.e. doesn't rely on anything from config.h. Additionally it can't be tested that the main binary doesn't rely on anything from the tests, i.e. something (a dependency search) was wrongly added in the test/example folder but should have been in the main instead.

But given that there is only a single header this can be verified manually and your approach is quite easy to do.

evpobr:
> The downside I see is that this way it can't be tested whether the main header is "clean", i.e. doesn't rely on anything from config.h.

This is too much even for a paranoid person like me :smile:

PR #81, opened 2019-07-29
Require C99 compiler

### LINKED ISSUE #77: BUG: Use C99 flexible arrays instead of type[1] trailing member
Related to #68 as it requires C99 (although C99 features are already used)

C99 flexible arrays are used at some points, e.g.: https://github.com/erikd/libsamplerate/blob/2647fd9bac5f11f24b5f1ee55806671f81520f66/src/src_sinc.c#L57

At other the idiom of having a trailing `float[1]` member is used: https://github.com/erikd/libsamplerate/blob/2647fd9bac5f11f24b5f1ee55806671f81520f66/src/src_linear.c#L34

A SO answer highly recommends using the C99 feature: https://stackoverflow.com/a/247040/1930508 with a comment that using `float[1]` may produce incorrect code: https://lkml.org/lkml/2015/2/18/407

Furthermore to much memory is allocated (1 additional float): https://github.com/erikd/libsamplerate/blob/2647fd9bac5f11f24b5f1ee55806671f81520f66/src/src_linear.c#L174

Bottom line: 
- Require C99 compiler
- Use flexible arrays instead of trailing `float[1]`

Flamefire:
Already required for configure builds, now also for CMake builds.
Simplifies some code as e.g. lrint and stdint.h are always available.
Closes #77

Note: Always links to libm (if found) as required by the C standard when using functions from math.h. If those are inlined the linker will discard anything from libm which isn't used, even everything. So there is no downside. The previous attempt checking IF linking it is required was brittle, because multi-config generators might require it for Debug builds but not Release builds causing compile failures.


Flamefire:
Rebased on master. @erikd Any more issues with this?

erikd:
Has MSVC  caught up to C99 yet? For the longest time (at least 5 years since I last checked) it hadn't.

Flamefire:
It at least supports all features used (VAM, stdint.h, lrint, ...) You can double-check if you merge #82 which uses VS 2017 on travis.

PR #86, opened 2019-08-09
 Remove shift of signed int in src_float_to_short_array 

Flamefire:
1. Scale by -SHORT_MIN
2. Clip if float is outside range of short
3. Otherwise round and truncate (guaranteed to fit to short)

Uses float instead of double as float can fit all 16 bit values.

Better alternative to approaches like #85, also adds test that show where e.g. https://github.com/janstary/libsamplerate/commit/ec0776030f02a6ef84f034c56e23e801a1f6be26 would fail.

Omits the clipping optimization which is normally disabled on x64 due to `long` being 64bits.


Flamefire:
After this was merged: Would it make sense to do the same for `src_float_to_int_array`? While it does not have the saved instruction when clipping optimization is disabled it would still look cleaner/more readable and avoid another configure check.

Sidenote: Unfortunately using `lrint` is a pessimization on remotely modern CPUs.:

- `lrint` is required to handle overflow errors by setting ERRNO: http://www.cplusplus.com/reference/cmath/lrint/
- without `-fno-math-errno` `lrint` won't get inlined but compiled to a library call due to the above making it slow
- "modern" CPUs support the truncating rounding conversion of float/double->int in a single instruction

Docu: http://www.cplusplus.com/reference/cmath/lrint/   
Experiment: https://godbolt.org/z/l77Mvt

erikd:
How do we get the "truncating rounding conversion of float/double->int in a single instruction"?

Flamefire:
Simply use a plain cast. See the godbolt link for resulting assembly in comparison

erikd:
Does a plain cast have the same rounding behavior as `lrint`? If it does not, its not a replacement.

Flamefire:
Yes and no. A cast is truncating so the same as "round to zero". So if the rounding mode is set to `FE_TOWARDZERO` then `lrint` does the same. Otherwise a "round to nearest" can be emulated by `(x>0) ? x+.5f : x-.5f;` which clang optimizes to a branch-less version on x86_64. Note however that this "round to nearest" is not guaranteed when you use `lrint`.

A good read with comparisons on different archs: https://stackoverflow.com/a/37624488/1930508

PR #87, opened 2019-08-23
Add failing test with chunked input

Flamefire:
There is a bug when using `src_process` with sufficiently small chunks of input data. Although the result of passing the input data in chunks should be exactly the same as when passing them as 1 block the test shows that this is not the case and NO output is generated.

I'm unsure how to fix this, so this only adds the failing test for further investigation.

### DIFF
--- a/tests/chunked_test.c
+++ b/tests/chunked_test.c
@@ -1,0 +1,105 @@
+/*
+** Copyright (c) 2002-2016, Erik de Castro Lopo <erikd@mega-nerd.com>
+** All rights reserved.
+**
+** This code is released under 2-clause BSD license. Please see the
+** file at : https://github.com/erikd/libsamplerate/blob/master/COPYING
+*/
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <samplerate.h>
+#include "util.h"
+
+void chunked_input_test(int converter, int channels, long chunkSize, double ratio)
+{
+	printf("chunked_input_test(converter=%d, channels=%d, chunkSize=%ld, ratio=%g)\n",
+		converter, channels, chunkSize, ratio);
+
+	const long NUM_FRAMES = 30*256;
+	int error;
+	
+	float *input = calloc(NUM_FRAMES * channels, sizeof(float));
+	TEST(input);
+	float *output = calloc(NUM_FRAMES * channels, sizeof(float));
+	TEST(output);
+	float *chunkedOutput = calloc(NUM_FRAMES * channels, sizeof(float));
+	TEST(chunkedOutput);
+
+	for (int ch = 0 ; ch < channels ; ch++)
+	{	double freq = 0.0111 ;
+		gen_windowed_sines (1, &freq, 1.0, output + ch * NUM_FRAMES, NUM_FRAMES) ;
+	}
+	interleave_data(output, input, NUM_FRAMES, channels);
+
+	// Calculate in 1 pass
+	SRC_DATA onePassData;
+	memset (&onePassData, 0, sizeof (onePassData)) ;
+	onePassData.data_in = input;
+	onePassData.data_out = output;
+	onePassData.input_frames = NUM_FRAMES;
+	onePassData.output_frames = NUM_FRAMES;
+	onePassData.src_ratio = ratio;
+	CHECKED_CALL(src_simple(&onePassData, converter, channels));
+
+	SRC_STATE *state;
+	if ((state = src_new (converter, channels, &error)) == NULL)
+	{
+		printf ("\n\nLine %d : src_new() failed : %s\n\n", __LINE__, src_strerror (error)) ;
+		exit (1) ;
+	}
+	// Calculate in chunks
+	SRC_DATA data;
+	memset (&data, 0, sizeof (data)) ;
+	data.data_in = input;
+	data.data_out = chunkedOutput;
+	data.output_frames = NUM_FRAMES;
+	data.src_ratio = ratio;
+
+	long input_frames = NUM_FRAMES;
+	long input_frames_used = 0;
+	long output_frames_gen = 0;
+
+	while(input_frames && data.output_frames)
+	{
+		data.input_frames = MIN(input_frames, chunkSize);
+		data.end_of_input = data.input_frames == input_frames;
+
+		CHECKED_CALL(src_process (state, &data));
+		input_frames_used += data.input_frames_used;
+		output_frames_gen += data.output_frames_gen;
+
+		data.input_frames -= data.input_frames_used;
+		data.data_in += data.input_frames_used * channels;
+		data.output_frames -= data.output_frames_gen;
+		data.data_out += data.output_frames_gen * channels;
+
+		input_frames -= data.input_frames_used;
+	}
+	state = src_delete (state);
+
+	TEST(onePassData.output_frames_gen == output_frames_gen);
+	for(long i = 0; i < output_frames_gen * channels; i++)
+	{
+		if(output[i] != chunkedOutput[i])
+			TEST(output[i] == chunkedOutput[i]);
+	}
+	
+	free(input);
+	free(output);
+	free(chunkedOutput);
+}
+
+int
+main (void)
+{
+	int CONVERTERS[] = {SRC_SINC_FASTEST, SRC_ZERO_ORDER_HOLD, SRC_LINEAR};
+	double RATIOS[] = {1., 2., 256., 1./2., 1./256.};
+
+	for(int iConverter = 0; iConverter < ARRAY_LEN(CONVERTERS); iConverter++)
+		for(int iRatio=0; iRatio < ARRAY_LEN(RATIOS); iRatio++)
+			chunked_input_test(CONVERTERS[iConverter], 1, 128, RATIOS[iRatio]);
+	return 0 ;
+}
--- a/tests/util.h
+++ b/tests/util.h
@@ -29,6 +29,19 @@ double calculate_snr (float *data, int len, int expected_peaks) ;
 
 const char * get_cpu_name (void) ;
 
+/** Assertions for tests */
+#define CHECKED_CALL(call) \
+	if ((error = call)) \
+	{	printf ("\n\nSRC call failed on Line %d : %s -> %s\n\n", __LINE__, #call, src_strerror (error)) ; \
+		exit (1) ; \
+	}
+
+#define TEST(condition) \
+	if (!(condition)) \
+	{	printf ("Condition failed on Line %d : %s\n\n", __LINE__, #condition) ; \
+		exit (1) ; \
+	}
+
 #if OS_IS_WIN32
 /*
 **	Extra Win32 hacks.


erikd:
> sufficiently small chunks of input data. 

Libsamplerate is designed for use on input streams that are (effectively) infinite in length. Using libsamplerate on short chunks of audio will have issues like the ones you describe here.

I am totally uninterested in whatever problems may arise from processing sub-second chunks of audio or in fact any artifical scenario that deviates too far from the purpose of the library.

Flamefire:
> Libsamplerate is designed for use on input streams that are (effectively) infinite in length. Using libsamplerate on short chunks of audio will have issues like the ones you describe here.

Have you even tried the test?

According to the documentation `src_process` can be used to process an input (file/stream/...) in chunks. The only mentioned pitfall is not providing enough output buffer (length) in which case the internal buffering will eventually be filled (but this should be reported by `input_samples_consumed = 0`) Having a workflow like:
- open input and output file
- allocate buffers for input and output with a given chunksize to not read GBs of data at once
- call `src_process` on the current chunk
- write any output
- advance buffers by what is reported by `src_process`
- repeat

This should work and produce the same result as invoking `src_simple` with the complete input file and a sufficiently large output buffer.

This is not the case! During developing the varispeed_test I've seen that passing data in 128Byte chunks into src_process does not produce any/enough output if the ratio is high enough. 

However it seems I cannot reproduce this in this test case. **Instead** it shows that `src_simple` produces not enough data. `onePassData.output_frames_gen` is zero when the ratio is low enough. The reason is that `half_filter_chan_len` gets large but `prepare_data` initially just copies a few zeros and no input data. This makes `samples_in_hand <= half_filter_chan_len` succeed aborting the loop without producing any data. So either `prepare_data` should be fixed to not only add zeros but also consume (some) input on initial call or `src_simple` should call `src_process` in a loop

So this is a real problem on a normal use case with (at least) `src_simple`

erikd:
> Have you even tried the test?

Yes, I tried it. The failure is:
```
Condition failed on Line 84 : onePassData.output_frames_gen == output_frames_gen
```
First of all, that failure message is close to useless because it it says nothing about what the actual values are.

Replacing that macro with something more sensible, and the error becomes:
```
chunked_input_test(converter=2, channels=1, chunkSize=128, ratio=1)
chunked_input_test(converter=2, channels=1, chunkSize=128, ratio=2)
chunked_input_test(converter=2, channels=1, chunkSize=128, ratio=256)
chunked_input_test(converter=2, channels=1, chunkSize=128, ratio=0.5)
chunked_input_test(converter=2, channels=1, chunkSize=128, ratio=0.00390625)
Line 85 : 0 != 178
```
For very small ratios, and small chunks of audio with the SINC converters, its it perfectly reasonable for the output length to be zero. I took one look at that and decide to move on. My time is scarce and *very* valuable to me.

I repeat libsamplerate is designed to be used for effectively infinite streams of audio. What happens with sub second chunks is irrelevant.

> (but this should be reported by input_samples_consumed = 0) 

That shows a fundamental misunderstanding of how this library and its API works. It is perfectly valid for input to be consumed without output being generated. The SINC converters have a transport delay which means they need to consume a certain number of input samples before any output can be generated.

I have spend an *incredible* amount of my very valuable and very scarce time responding to your comments and PRs over the last couple of weeks. I have had some good things come out this sunk time, but IMO I have not yet broken even. Furthermore, what I thought was arrogance and pushiness on your part may simply have been you not considering how much of my time you were consuming and how scarce that time is.

Flamefire:
Ok summary: `src_simple` does not work for small ratios
- The test does the following:
  - Convert N samples of data via `src_simple`
  - Convert N samples of (the same) data via `src_process` using C bytes at once
  - Compare the results to be equal
  - Real world use case: Processing a file via an arbitrarily sized buffer
- Bug found:
  - `src_simple` does **never** produce any output when `src_ratio` is small (ca. `<=1/215`)
- Reason: 
  - `prepare_data` does not consume enough input on first call
  - For low ratios this is not enough to produce any output which exits the loop and function without calling `prepare_data` again
  - As the amount of data copied is bounded by `b_len` I thin this might be wrong, especially as there is a `FIXME` comment above the calculation

Again: This does NOT depend on the amount of input but ONLY on the ratio. There is a constant `NUM_FRAMES` that you could increase to any value to verify this. I set it to a low value which shows the same behavior to speed up the test.

> Replacing that macro with something more sensible, and the error becomes:

Great. I failed to create a macro like `ASSERT_EQUAL(lhs, rhs)` in C due to missing generic string conversion or function overloading. This is why I went with the most generic macro for the tests.

> What happens with sub second chunks is irrelevant.

See above: This happens for ANY sized input passed to `src_simple` when the ratio is low. Sorry for initially assuming `src_process` and chunk sizes were the issue. This was what I initially observed in another test and thought would be reproduced by this. Instead it showed an issue with `src_simple`

> That shows a fundamental misunderstanding of how this library and its API works.

You misunderstood my point. I'm aware of the transport delay, but that doesn't matter for this issue.

erikd:
> Ok summary: src_simple does not work for small ratios

Small ratios for long pieces of audio? I 100,000 frames or more? Eg, a typical pop song is 3 minutes in length, which is 180 seconds, which with CD quality is 7938000 frames. If `src_simple` doesn't work for that, then yes, it *may* be work fixing (ignoring the fact that extreme ratios don't have that many real applications).

Flamefire:
> Small ratios for long pieces of audio? Like 100,000 frames or more? 

Yes: https://github.com/Flamefire/libsamplerate/commit/d3c26fb8d01818cd7385c4cc6fc605e3e5bb73cf   
See result: https://travis-ci.com/Flamefire/libsamplerate/jobs/227997242

> ignoring the fact that extreme ratios don't have that many real applications

Sure. I wouldn't mind if these ratios simply fail with an error. But not silently and depending on which method is used although they should be equivalent. Especially if the "simple" method fails where a user expects that the tricky stuff (e.g. calling src_process until no more output is produced) is handled in the library.

erikd:
Pointing me a travis output which does not contain the exact specific details of what is being tested is a flagrant waste of my time. Even the output:
```
chunked_input_test (converter=2, channels=1, chunkSize=128, ratio=0.00390625)
Condition failed on Line 83 : onePassData.output_frames_gen == output_frames_gen
```
is useless because I can't see the size of the input data, nor can I see whether you are using `src_simple` or `src_process`.

Flamefire:
So what do you need? I pointed you to a very specific commit on top of this PR which increases the number of frames to 100000 as requested. You already tried this PR (with smaller number of frames) and seen that `onePassData.output_frames_gen == 0`. The same failure happens with the increased number of frames. 

I now added a version of `ASSERT_EQUAL` and enabled failing test output:   
[CI now shows](https://travis-ci.com/Flamefire/libsamplerate/jobs/227999007)
```
 3/17 Test  #3: chunked_test .....................***Failed    0.63 sec
chunked_input_test (converter=2, channels=1, chunkSize=128, ratio=1)
chunked_input_test (converter=2, channels=1, chunkSize=128, ratio=2)
chunked_input_test (converter=2, channels=1, chunkSize=128, ratio=256)
chunked_input_test (converter=2, channels=1, chunkSize=128, ratio=0.5)
chunked_input_test (converter=2, channels=1, chunkSize=128, ratio=0.00390625)
Condition onePassData.output_frames_gen == output_frames_gen failed on Line 83
onePassData.output_frames_gen == 0
output_frames_gen == 382
```

This is still the same test we were talking about in this PR. `onePassData.output_frames_gen` is filled by `src_simple`: https://github.com/Flamefire/libsamplerate/blob/68035a57b4f3f26d0161f01f62cc51146f243b67/tests/chunked_test.c#L45

I still think the comparison between `src_simple` and `src_process` output is valuable. As a byproduct it shows the failure of `src_simple`. If required a test of `src_simple` can be factored out from the existing test easily. I just don't know what to compare the data against in that case besides that there is ANY data.

erikd:
That `onePassData.output_frames_gen == 0` with input frames being `100000` and the conversion ratio being `0.00390625` is not surprising to me. `100000 * 0.00390625 => 390.625` which is almost certainly the transport delay in samples of that converter.

If you increase the ratio or increase the number of frames (while keeping the other constant),  `output_frames_gen` will no longer be zero. This is not a big. This is the behaviour inherent to this kind of sample rate converter.

erikd:
I have said it before, I will say it again; libsamplerate is designed to used on (effectively) infinite stream of audio.

The function `src_simple` was added a convenience function. It is possible for the user to call it with parameters which result in no output. I would accept a patch that returns an error for parameter which result in no output as long as that does not impact the performance of `src_process`.

Flamefire:
From my debugging `prepare_data` does not consume more input if the input-frames are increased for that ratio as the amount of data "generated" by `prepare_data` is bounded by `b_len`.

But ok, then 2 questions:
- The chunked `src_process` version with that ratio produces the first output sample after 4992 input frames are used. Doesn't that mean the transport delay is <= 4992?
- My claim is that the bug occurs for ratios ~ <=1/215, so only option is to increase the number of frames as the ratio is valid inside the library. What number of frames do you propose? Increasing it to 10^6 produces 3897 frames for src_process and still zero frames for src_simple. Note that this is consistent: `10^6*1/256=3906` which is close to the 3897 frames produced by src_process. Same for 10^5

>  libsamplerate is designed to used on (effectively) infinite stream of audio.

See above: please provide an example of an "infinite stream" to pass into src_simple with a ratio <= 1/215 where any output is produced.

Please can you double check the calculations at https://github.com/erikd/libsamplerate/blob/1b3d5b1114b6cac1baed0a1a24ea8bb451cfc1b1/src/src_sinc.c#L215 with respect to their use at https://github.com/erikd/libsamplerate/blob/1b3d5b1114b6cac1baed0a1a24ea8bb451cfc1b1/src/src_sinc.c#L1178 and https://github.com/erikd/libsamplerate/blob/1b3d5b1114b6cac1baed0a1a24ea8bb451cfc1b1/src/src_sinc.c#L386 through https://github.com/erikd/libsamplerate/blob/1b3d5b1114b6cac1baed0a1a24ea8bb451cfc1b1/src/src_sinc.c#L1199 and https://github.com/erikd/libsamplerate/blob/1b3d5b1114b6cac1baed0a1a24ea8bb451cfc1b1/src/src_sinc.c#L1208)

erikd:
I don't have time for this. If there is a bug fix it.

Flamefire:
How would anyone but you be able to understand these calculations?

I found a bug, analyzed it and pointed you to the most likely source code position which contains some calculations which seem to be wrong. I even provided a test case to reproduce and changed it according to your comments. What more do you expect from someone not deeply familiar with that code?

erikd:
>  What more do you expect from someone not deeply familiar with that code?

I wrote this code 15 years ago. At that time I probably was deeply familiar with it, but now< i am probably no more familiar with it than you.

You *think* you have found a bug. I am not going to look at what you *think* may be a bug, but I will look at a proposed *fix* to this bug. Viewing the fix will help me figure out if this actually *is* a bug.

Issue #5, opened 2016-09-27
Potential Bad memory access bug


hnand:
Hi Erik,

I have written a self-contained test to display the problem we are seeing in our project that uses libsamplerate to do sample rate conversions. I have attached the entire libsamplerate directory along with my changes here. Additionally, I have also written up a brief description of the issue and my test below. 
I would like to understand if this is indeed a potential bug or an improper use of the library. Looking forward to your thoughts.

Thanks,
Hari

**Description:**

Our project code follows the model in callback_test.c (ie: setup a callback using src_callback_new() with converter = SRC_SINC_FASTEST, numChannels = 32; calls to src_callback_read() till the requested # of samples are read).

We have discovered an intermittent issue where the audio data we receive after a call to src_callback_read() contains Nan values in the output buffer at a valid index (ie: within the total read samples returned by src_callback_read()). We traced this down to the calc_output_multi() function which is called by  sinc_multichan_vari_process() in src_sinc.c. Specifically, in the scenarios where the issue happens, the data_index that calc_output_multi() calculates when applying the left half of the filter evaluates to a negative value (which leads to accessing uninitialized/unowned memory).

 To simplify this, I have written a short self-contained test (called filterdata_integrity_test.c) under the tests directory. This test is similar to callback_test.c, ie: it sets up a callback using src_callback_new() with converter = SRC_SINC_FASTEST, numChannels = 2 (for simplicity); calls to src_callback_read() till the requested # of samples are read). The notable differences from my test and callback_test.c are:
i)                     Setting the src_ratio to a different value on every callback (by calling src_set_ratio()) -> this replicates the scenario in our project when the issue occurs
ii)                   Adding integrity checks on the audio data that is generated in the output buffer
All of the relevant changes also have comments next to them.

Note: I have also made an un-invasive change to src/src_sinc.c at line 437 to exit(1) whenever the data_index hits a negative value in calc_output_stereo() (since this is the only way I have found to catch the issue when it happens). I hit this exit every time I run the filterdata_integrity_test.

[libsamplerate-0.1.8.zip](https://github.com/erikd/libsamplerate/files/496420/libsamplerate-0.1.8.zip)

erikd:
I'm managed to confirm this. Working on a fix, but it may only land when I get back from holidays next week.

erikd:
The documentation suggests that setting the ratio using `set_src_ratio` is valid so this is definitely a bug in handling of some edge case.

hnand:
[...]

PR #89, opened 2019-08-23
 Add test shwoing memory corruption when using src_set_ratio

Flamefire:
As described in https://github.com/erikd/libsamplerate/issues/5#issuecomment-519166957 there is a serious bug when using a variable ratio that decreases. In the tests this results in either NaNs in the output or a crash when
accessing invalid memory (or an error when using ASAN) due to `data_index` being negative.

This gets more serious the more channels are used and the higher the ratio difference is as that makes the index more negative causing it first to point to data inside the `SINC_FILTER` struct, then in front of it (which might not even be allocated)

This adds a test for reproducing this, no fix contained so CI failure is expected

### DIFF
--- a/tests/util.h
+++ b/tests/util.h
@@ -29,6 +29,19 @@ double calculate_snr (float *data, int len, int expected_peaks) ;
 
 const char * get_cpu_name (void) ;
 
+/** Assertions for tests */
+#define CHECKED_CALL(call) \
+	if ((error = call)) \
+	{	printf ("\n\nSRC call failed on Line %d : %s -> %s\n\n", __LINE__, #call, src_strerror (error)) ; \
+		exit (1) ; \
+	}
+
+#define TEST(condition) \
+	if (!(condition)) \
+	{	printf ("Condition failed on Line %d : %s\n\n", __LINE__, #condition) ; \
+		exit (1) ; \
+	}
+
 #if OS_IS_WIN32
 /*
 **	Extra Win32 hacks.
--- a/tests/varispeed_test.c
+++ b/tests/varispeed_test.c
@@ -26,10 +26,14 @@
 #define	BUFFER_LEN		(1 << 16)
 
 static void varispeed_test (int converter, double target_snr) ;
+static void set_ratio_test (int converter, int channels, double initial_ratio, double second_ratio) ;
 
 int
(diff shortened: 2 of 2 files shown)


erikd:
This was not a memory corruption error because the error was due to a read not a write. It would be more correct to call this an out-of-bounds read error.

I have fixed this in #91 and added an extensively reworked version of your test.

Flamefire:
> It would be more correct to call this an out-of-bounds read error.

Agreed. Depending on how the ratio is gotten this can still be a DoS vulnerability as it crashes the program.

> an extensively reworked version of your test.

You should have just worked with me to improve this version. I commented on many issues on your (mostly copied) version which was not even properly tested as opposed to this.

erikd:
> You should have just worked with me

Really?

> I commented on many issues on your (mostly copied) version which was not even properly tested as opposed to this.

You really think your code was perfect?

Please feel free to fork this repo and never comment or raise a PR here again.

Flamefire:
> You really think your code was perfect?

Nothing is. But at least my test was verified (on CI) to reproduce the issue. Your test does not even reproduce the issue.   
If instead on changing many parts of the test you could have told me what you dislike or where an issue was and I could have explained the problem I wanted to solve with that part of the code or fix the issue if it was a bug/wrong. This is what reviews are for and how PRs are usually handled

erikd:
I really hope you are better at interacting with people IRL than you are over the internet.

Flamefire:
Probably. English is not my native language and written text is even more ambiguous then spoken words so what I've written might not be meant as it sounded. Particularly I did in no way wanted to be rude.

The sentence `You should have just worked with me to improve this version.` was meant as: "... this would have been easier then spending time rewriting it, then fixing issues on the merged code"   
On reviewing the changed test I've seen that it deviates in many important places from my intentions and does not even run into the bug it is intended to show making it almost useless. I spent quite some time to make sure my version does run into the issue it tries to show, your doesn't which is what I meant with the "not tested". Instead of rewriting it it would have been better to improve the existing test (which does run into the bug) in a dialogue with the author of the test.

The 2nd part was rather meant as: Now it's to late. However I check for and commented  on the issues I found and written next to your code for how to fix/improve the version that is now used.

PR #91, opened 2019-08-24
Fix an out-of-bounds read error

erikd:
The bad data that is read is audio data in floating point format that is immediately multiplied by a compile time constant value (read from a very large table of floating point values) and then added to an accumulator.

As such, this is not an exploitable bug.

### DIFF
--- a/tests/varispeed_test.c
+++ b/tests/varispeed_test.c
@@ -20,12 +20,14 @@
 #if HAVE_FFTW3
 #include <fftw3.h>
 #else
-#define fftw_cleanup()
+#define fftw_cleanup ()
 #endif
 
 #define	BUFFER_LEN		(1 << 16)
 
 static void varispeed_test (int converter, double target_snr) ;
+static void varispeed_hammer_test (int converter) ;
+static void set_ratio_test (int converter, int channels, double initial_ratio, double second_ratio) ;
 
 int
 main (void)
@@ -33,12 +35,18 @@ main (void)
 	puts ("") ;
 	printf ("    Zero Order Hold interpolator    : ") ;
 	varispeed_test (SRC_ZERO_ORDER_HOLD, 10.0) ;
+	varispeed_hammer_test (SRC_ZERO_ORDER_HOLD) ;
+	puts ("ok") ;
 
 	printf ("    Linear interpolator             : ") ;
 	varispeed_test (SRC_LINEAR, 10.0) ;
+	varispeed_hammer_test (SRC_LINEAR) ;
+	puts ("ok") ;
 
 	printf ("    Sinc interpolator               : ") ;
 	varispeed_test (SRC_SINC_FASTEST, 115.0) ;
+	varispeed_hammer_test (SRC_SINC_FASTEST) ;
+	puts ("ok") ;
 
 	fftw_cleanup () ;
 	puts ("") ;
@@ -65,7 +73,7 @@ varispeed_test (int converter, double target_snr)
 
 	/* Perform sample rate conversion. */
 	if ((src_state = src_new (converter, 1, &error)) == NULL)
-	{	printf ("\n\nLine %d : src_new() failed : %s\n\n", __LINE__, src_strerror (error)) ;
+	{	printf ("\n\nLine %d : src_new () failed : %s\n\n", __LINE__, src_strerror (error)) ;
 		exit (1) ;
 		} ;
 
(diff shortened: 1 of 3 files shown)


tests/varispeed_test.c
```diff
@@ -144,8 +152,101 @@ varispeed_test (int converter, double target_snr)
...
 } /* varispeed_test */
 
+static void
+varispeed_hammer_test (int converter)
+{	double ratios [] = { 1.0, 0.01, 20 } ;
+	int chan, r1, r2 ;
+
+	for (chan = 1 ; chan <= 9 ; chan ++)
+		for (r1 = 0 ; r1 < ARRAY_LEN (ratios) ; r1++)
+			for (r2 = 0 ; r1 != r2 && r2 < ARRAY_LEN (ratios) ; r2 ++)
```
Flamefire: Condition in the loop header does abort the loop instead of continuing it which results in not testing all combinations as intended

tests/varispeed_test.c
```diff
@@ -144,8 +152,101 @@ varispeed_test (int converter, double target_snr)
...
+	/* Now switch to second_ratio and process the remaining. */
+	src_data.src_ratio = second_ratio ;
+
+	while (1)
+	{	if ((error = src_process (src_state, &src_data)))
+		{	printf ("\n\nLine %d : %s : %s\n\n", __LINE__, details, src_strerror (error)) ;
+			exit (1) ;
+			} ;
+
+		if (src_data.end_of_input && src_data.output_frames_gen == 0)
```
Flamefire: Omitting the condition of `output_len <= current_out` leads to an infinite loop on e.g. `first_ratio==second_ratio==20`

tests/varispeed_test.c
```diff
@@ -144,8 +152,101 @@ varispeed_test (int converter, double target_snr)
...
 
+static void
+varispeed_hammer_test (int converter)
+{	double ratios [] = { 1.0, 0.01, 20 } ;
+	int chan, r1, r2 ;
+
+	for (chan = 1 ; chan <= 9 ; chan ++)
+		for (r1 = 0 ; r1 < ARRAY_LEN (ratios) ; r1++)
+			for (r2 = 0 ; r1 != r2 && r2 < ARRAY_LEN (ratios) ; r2 ++)
+				set_ratio_test (converter, 1, ratios [r1], ratios [r2]) ;
```
Flamefire: `chan` is not used executing the same test 9 times

tests/varispeed_test.c
```diff
@@ -144,8 +152,101 @@ varispeed_test (int converter, double target_snr)
...
 		} ;
 
-	puts ("ok") ;
-
 	return ;
 } /* varispeed_test */
 
+static void
+varispeed_hammer_test (int converter)
+{	double ratios [] = { 1.0, 0.01, 20 } ;
```
Flamefire: Missing corner case ratios. Adding `1./256.` shows a heap-user-after-free in ASAN. Also the initial test case as reported in #5 for ratios `0.09 -> 0.08` is missing (or more generally 2 ratios below 1 as e.g. `0.1 and 0.01` which was contained in my test)

tests/varispeed_test.c
```diff
@@ -144,8 +152,101 @@ varispeed_test (int converter, double target_snr)
...
+		src_data.data_out	+= src_data.output_frames_gen ;
+
+		src_data.input_frames	= input_len - current_in ;
+		src_data.output_frames	= output_len - current_out ;
+
+		src_data.end_of_input = (current_in >= input_len) ? 1 : 0 ;
+		} ;
+
+
+	for (k = 0 ; k < current_out * channels ; k ++)
```
Flamefire: Missing `ASSERT(current_out > 0)` which can cause a test success with no check of the output (in the case the buffers are to small to produce any output)

erikd:
I have fixed most of the issues above and find that without the `src/src_sinc.c` fix, the test program could still not reproduce the OOB read.

I then went back to your version in this and was immediately reminded of why I rewrote this test, I find the code in your test difficult to read/understand. Why are you using two `SRC_DATA` structs and what purpose does the `adjustData` function serve?

Flamefire:
If you are referring to https://github.com/erikd/libsamplerate/commit/1b3d5b1114b6cac1baed0a1a24ea8bb451cfc1b1: You are missing a `src_set_ratio` call below the comment `hard switch...`

Side notes:
- Using 2^15 frames input makes the test slow. anything above 256 should be fine (conversion ration 1/256 as minimum so 256 input frames should produce 1 output frame and producing any output is enough to trigger the bug, I chose some multiple of 256 to be safe [rounding ...])
- After first `src_process` you don't update `src_data` so if output is produced, you will overwrite it
- I would not use the `max_loop_count` loop as this makes it hard for someone looking how to use the library to understand the break condition that should be used in real-world-apps (coming from "tests are docu")

> Why are you using two SRC_DATA structs and what purpose does the adjustData function serve?

This would have something I could have worked with and improved the test. `adjustData` was used to avoid the code duplication after `src_process` (see 2nd point above where you missed that piece of code I put into the function)

PR #94, opened 2019-08-27
Speed improvements

Flamefire:
In #82 I noticed that 'Duff's device' for manually unrolled loops is wrongly implemented leading to a heavy degradation in speed. Additionally compilers are so good in optimizing that manually unrolled loops are often slower than compiler optimized ones. See https://github.com/erikd/libsamplerate/pull/82#discussion_r312931906 and Benchmark at http://quick-bench.com/znRud0mQhKSWDhgJns_5eYP8_wk

Additionally the fix for #5 in https://github.com/erikd/libsamplerate/commit/c10c8191f21dd12c1bea2d9b9a68536f7cb0491c introduced an `if` in one of the main worker loops although that can be easily moved out with same semantic.

Changes in this PR:
- Add assertions to check for valid indices (optimized out in release builds, so no speed penalty there. Very useful for tests in cases ASAN cannot find or is not used) before changing anything
- Test more channels in `multichan_throughput_test --best-of <n>` and make its output easier parseable
- Replace manually unrolled loops (especially the buggy Duff's device) with regular loops
- Move the `if(data_index < 0)` handling out of the loop (which requires a change from `do-while` to `while` or something like a GOTO)

Measured performance difference on 2 machines. Speedup is up to 2.2x for the loop-transformation and 1.2x for the if-moving for `--best-of 5`. Got some spurious slow-downs though in  the 5%  range which disappeared when running the test again. So probably some kernel-interupt or so.

Graphs:
bench1
bench2
bench3
bench4

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -223,7 +223,7 @@ dnl ============================================================================
 dnl  Compiler stuff.
 
 AS_IF([test "x$enable_flags_setting" = "xyes"], [
-		AX_APPEND_COMPILE_FLAGS([-DHAVE_CONFIG_H -O2 -pipe], [CFLAGS])
+		AX_APPEND_COMPILE_FLAGS([-DHAVE_CONFIG_H -O2 -pipe -DNDEBUG], [CFLAGS])
 
 		AS_CASE([${host_os}],
 			[darwin*], [
--- a/src/src_sinc.c
+++ b/src/src_sinc.c
@@ -903,10 +903,9 @@ sinc_hex_vari_process (SRC_PRIVATE *psrc, SRC_DATA *data)
 static inline void
 calc_output_multi (SINC_FILTER *filter, increment_t increment, increment_t start_filter_index, int channels, double scale, float * output)
 {	double		fraction, icoeff ;
-	/* The following line is 1999 ISO Standard C. If your compiler complains, get a better compiler. */
 	double		*left, *right ;
 	increment_t	filter_index, max_filter_index ;
-	int			data_index, coeff_count, indx, ch ;
+	int			data_index, coeff_count, indx ;
 
 	left = filter->left_calc ;
 	right = filter->right_calc ;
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -5,13 +5,15 @@ dist: xenial
...
-        - sudo add-apt-repository ppa:ondrej/autotools -y # automake 1.15
-        - sudo add-apt-repository ppa:cz.nic-labs/knot-dns -y # pkg-config 0.29.2
-        - sudo apt-get update
-        - sudo apt-get install -y autoconf automake-1.15 pkg-config m4
+      addons:
+        apt:
+          packages: [libsndfile-dev, libfftw3-dev, libasound2-dev, autoconf, automake-1.15, pkg-config, m4]
+        sources:
+          - sourceline 'ppa:ondrej/autotools' # automake 1.15
+          - sourceline 'ppa:cz.nic-labs/knot-dns' # pkg-config 0.29.2
```
Flow86, line 16: you changed this in #102 already - should not be part of this PR

Flamefire: This is true, but this PR predates #102 and this change was required to make the CI less flaky as it spuriously failed due to connection issues, which the apt-addon of travis workarounds by re-trying. I'll rebase and remove this once another PR is merged

Flamefire: As that change was merged I removed the commit from this PR now

evpobr:
Hi @Flamefire, are you still interested in this? This issue is good, but overcomplicated. 

Let's start from 500deda. Please, make one change per PR. Make a new pull request and I will merge it immediately, i promise :smile:

Flamefire:
Sure. How fine grained would you like the PRs? My suggestion would be:
- Assert & NDEBUG (first 2 commits): Those belong together or it would degrade performance by default
- Test change (3rd commit): Preparation to verify effect of this PR
- Speed optimizations (last 3 commits): This is one topic, so having 1 PR with 3 commits is usually fine (usually CI runs and you don't want to waste resources), but I'm open to splitting this further if required.

evpobr:
> Sure. How fine grained would you like the PRs? My suggestion would be:
> 
> * Assert & NDEBUG (first 2 commits): Those belong together or it would degrade performance by default
> * Test change (3rd commit): Preparation to verify effect of this PR
> * Speed optimizations (last 3 commits): This is one topic, so having 1 PR with 3 commits is usually fine (usually CI runs and you don't want to waste resources), but I'm open to splitting this further if required.

Then NDEBUG first. And how about [this macro](https://www.gnu.org/software/autoconf-archive/ax_check_enable_debug.html)? 

Note one commit per PR.

Flamefire:
Added 3 PRs for the first 3 commits. The others depend on those, so waiting for you to merge them first :)

evpobr:
> Added 3 PRs for the first 3 commits. The others depend on those, so waiting for you to merge them first :)

Merged!

Flamefire:
Ok, next 2 PRs filed. 1 remaining after those :)

evpobr:
@Flamefire , thanks. Closing this pull request.

Flamefire:
Thanks for addressing this. You may want to check my other PRs too, e.g. having Windows CI would be great. I've seen you did some fixes already that I suggested/made there regarding e.g. warnings on MSVC and also some CMake changes. So just check the individual commits and tell me which ones you'd like and I see that I resolve the conflicts.

PR #101, opened 2020-07-03
Fix CMake warning about case mismatch of sndfile

Flamefire:
CMake expects that the file/package name matches in casing. I.e. `find_package(Sndfile)` with a file `FindSndfile` needs `find_package_handle_standard_args(Sndfile ...`

The warning fixed is:

```
CMake Warning (dev) at C:/Program Files/CMake/share/cmake-3.17/Modules/FindPackageHandleStandardArgs.cmake:272 (message):
  The package name passed to `find_package_handle_standard_args` (SNDFILE)
  does not match the name of the calling package (Sndfile).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
```

### DIFF
--- a/cmake/FindSndfile.cmake
+++ b/cmake/FindSndfile.cmake
@@ -13,6 +13,6 @@ find_library(SNDFILE_LIBRARY NAMES sndfile
     HINTS $ENV{SNDFILE_ROOT}/lib)
 
 include(FindPackageHandleStandardArgs)
-find_package_handle_standard_args(SNDFILE DEFAULT_MSG SNDFILE_LIBRARY SNDFILE_INCLUDE_DIR)
+find_package_handle_standard_args(Sndfile DEFAULT_MSG SNDFILE_LIBRARY SNDFILE_INCLUDE_DIR)
 
 mark_as_advanced(SNDFILE_LIBRARY SNDFILE_INCLUDE_DIR)


cmake/FindSndfile.cmake
```diff
@@ -13,6 +13,6 @@ find_library(SNDFILE_LIBRARY NAMES sndfile
     HINTS $ENV{SNDFILE_ROOT}/lib)
 
 include(FindPackageHandleStandardArgs)
-find_package_handle_standard_args(SNDFILE DEFAULT_MSG SNDFILE_LIBRARY SNDFILE_INCLUDE_DIR)
+find_package_handle_standard_args(Sndfile DEFAULT_MSG SNDFILE_LIBRARY SNDFILE_INCLUDE_DIR)
```
Flow86, line 16: this will define a variable Sndfile_FOUND (variables are case sensitive!)

Flamefire: No. It will populate both of them according to the docs: https://cmake.org/cmake/help/v3.17/module/FindPackageHandleStandardArgs.html

> FOUND_VAR <result-var>
>
>    Obsolete. Specifies either `<PackageName>_FOUND` or `<PACKAGENAME>_FOUND` as the result variable. This exists only for compatibility with older versions of CMake and is now ignored. Result variables of both names are always set for compatibility.

evpobr:
Fixed in 09cb406.

PR #119, opened 2020-10-22
mingw build fixes, especially when cross-compiling. update Win32/*

sezero:


sezero:
Closing this. Better patch-set on the way.

PR #121, opened 2020-11-03
Move public header into include folder

Flamefire:
Extracted from #69

This is the minimal amount of work to not break anything. To make this actually useful the tests and examples should not include src or PROJECT_BINARY_DIR which serves as a test that the public header is self contained
For this the CMake files need to be split (see #69) and separate configs should be created which contain only what the tests/examples need. Some things can be removed, e.g. `HAVE_STDBOOL_H` is assumed to be defined as otherwise `true`, `false` and `bool` are not available and the build would break.

Note: Mixing automake and cmake for testing doesn't work because automake pollutes the source folder with e.g. the generated config.h which will then be picked up when building with cmake and as it is explicitly ignored it doesn't show up in `git status` (one of the reasons I'd remove the autotools based build)


CMakeLists.txt
```diff
@@ -198,116 +198,105 @@ if(BUILD_SHARED_LIBS)
     endif()
 endif()
 
+add_library(samplerate_config INTERFACE)
+# To allow src_config.h and config.h to be included
+target_include_directories(samplerate_config INTERFACE src ${PROJECT_BINARY_DIR})
+add_library(samplerate::config ALIAS samplerate_config)
```
evpobr, line 204: No better than `target_include_directories`.  Entities should not be multiplied without necessity.

evpobr (changes requested):

CMakeLists.txt
Flamefire: IMO it is better: If you compare it to a programming language this is the equivalent of a named function. So instead of shoving some folders in there (which BTW needs an extra command while you already have a target_link_libraries) you link to a named entity, in this case one meant to give you access to the configuration. The comment makes the intention clear, repeating that on every use would be noisy. And if that folder ever need to change there is a single place for it. So again this can be though of as `string getConfigIncludeFolder(){ ... }`
-> Diff shows that: 13 LOC less after this change which IMO is better for maintainability.

evpobr: > IMO it is better: If you compare it to a programming language this is the equivalent of a named function. So instead of shoving some folders in there (which BTW needs an extra command while you already have a target_link_libraries) you link to a named entity, in this case one meant to give you access to the configuration. The comment makes the intention clear, repeating that on every use would be noisy. And if that folder ever need to change there is a single place for it. So again this can be though of as string getConfigIncludeFolder(){ ... }

Do not forget that other people will also look into the project, who are not particularly interested in understanding what kind of library it is and why it is needed. On the other hand, target_include_directories() is clear to anyone, even with a basic knowledge of CMake. So my verdict is no.

> -> Diff shows that: 13 LOC less after this change which IMO is better for maintainability.

Compact code is also not particularly important, we are not writing in Assembler here.

PR #124, opened 2020-11-12
src_sinc.c: change SHIFT_BITS sanity check into compile time assert

sezero:


sezero:
I have been sitting on the following patch for quite some time,
don't know if it's acceptable for mainstream:
https://github.com/sezero/libsamplerate/commit/1ed2bb846cf6223ca0c41bcef458aef530fcf042
If it is, I can make it into a PR here.

evpobr:
> I have been sitting on the following patch for quite some time,
> don't know if it's acceptable for mainstream:
> [sezero@1ed2bb8](https://github.com/sezero/libsamplerate/commit/1ed2bb846cf6223ca0c41bcef458aef530fcf042)
> If it is, I can make it into a PR here.

Sorry, C99 only, thinking about C11.

PR #126, opened 2020-11-16
add NuttX OS build hooks

jerpelea:
Add the needed hooks for the NuttX OS build system.
NuttX OS uses Kconfig to set the needed configuration options.

Signed-off-by: Alin Jerpelea <alin.jerpelea@sony.com>


jerpelea:
@evpobr

evpobr:
Does it really have to be here?

jerpelea:
@evpobr I can try to find another way if you would like it to be moved on the other side

PR #134, opened 2020-12-11
several build updates

sezero:


evpobr (approved):

examples/audio_out.c
```diff
@@ -6,21 +6,25 @@
...
 */
 
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#ifdef HAVE_UNISTD_H
```
SoapGentoo, line 16: I assume this is for Windows?

examples/varispeed-play.c
```diff
@@ -6,16 +6,18 @@
 ** file at : https://github.com/libsndfile/libsamplerate/blob/master/COPYING
 */
 
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
```
SoapGentoo, line 11: thank you, `config.h` should always be the first include in every `.c` file

src/src_zoh.c
```diff
@@ -6,15 +6,15 @@
 ** file at : https://github.com/libsndfile/libsamplerate/blob/master/COPYING
 */
 
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
```
SoapGentoo, line 11: make it consistent: Add an empty line after this block in every file

tests/clone_test.c
```diff
@@ -28,7 +32,7 @@ clone_test (int converter)
 	SRC_STATE* src_state_cloned ;
 	SRC_DATA src_data, src_data_cloned ;
 
-	int error, frame, ch, index ;
+	int error, frame, ch, idx ;
```
SoapGentoo, line 35: why the change?

SoapGentoo (changes requested):

examples/audio_out.c
sezero: Practically, yes.

src/src_zoh.c
sezero: Done.

tests/clone_test.c
sezero: As mentioned in the commit message, it triggered -Wshadow and failed make distcheck on my setup.

SoapGentoo (approved):
@sezero looking good, could you squash related commits together? 10 commits seems like a lot for the changes here.

sezero:
> @sezero looking good,

Thanks.

> could you squash related commits together? 10 commits seems like a lot for the changes here.

Done

SoapGentoo:
@evpobr feel free to merge whenever you want. Also, given that libsamplerate doesn't use `autogen`, I think we should consider ditching autotools and moving to CMake only.

sezero:
>  given that libsamplerate doesn't use `autogen`,

Hm? `autogen` as in top-level `autogen.sh` in the source tree?

> I think we should consider ditching autotools and moving to CMake only.

Sad..

SoapGentoo:
@sezero autogen as in GNU Autogen (https://www.gnu.org/software/autogen/), it's a macro language we used in libsndfile for the test harnesses and has been a major compatibility nightmare between Unix and Windows...

> Sad..

oh are you in favour of the Autotools build system? On balance, I am usually too, because I don't like the accidental complexity CMake usually entails (All these NIH `Find*.cmake` scripts that could be simplified a lot using pkg-config). Let's shelf the idea then. It doesn't happen often that people in 2020 still cherish the Autotools...

sezero:
> oh are you in favour of the Autotools build system?

Oh yes.

> It doesn't happen often that people in 2020 still cherish the Autotools...

Well, I'm old :)

evpobr:
> @evpobr feel free to merge whenever you want. Also, given that libsamplerate doesn't use `autogen`, I think we should consider ditching autotools and moving to CMake only.

Actually you can merge this PR, you have the same rights in this repo as me.

SoapGentoo:
@evpobr I know, I was just putting you in charge ;)

evpobr:
> Also, given that libsamplerate doesn't use `autogen`, I think we should consider ditching autotools and moving to CMake only.

It's actually not very convenient to support two build systems. And if you choose between Autotools and CMake, I will of course choose the second one.

In fact, I was also thinking of removing Autotools from here. But first we will discuss everything and if we agree, we will first declare it deprecated and delete it after a few releases.

SoapGentoo:
@evpobr the main issue I have with the current CMake buildsystem is that it is too convoluted. If you're willing to switch to `pkg-config` as a first mechanism, and then fall back to guessing flags that would be fine. Upstreams may change in weird ways, and the only thing that remains steady is the output of `pkg-config --cflags <DEP>`, whereas `Find<DEP>.cmake` modules break so often in Gentoo, that they're a real pain to maintain.

evpobr:
I agree with that. I suggest we discuss this after the next release.

PR #141, opened 2021-01-22
Time resolution on GNU/Hurd is 10 msec not 1 usec as for other systems.

umlaeute:
this is an old patch we've been carrying in Debian for a couple of years.

it would be nice if it could be included upstream.


SoapGentoo:
@evpobr should be safe? `__GNU__` is only defined by glibc on GNU/Hurd, hence should be fine.

SoapGentoo:
@umlaeute in the mean time, please add a mention to `NEWS`

evpobr:
Yes, let's add it.

PR #142, opened 2021-01-22
calculate SHARED_VERSION_INFO from PACKAGE_VERSION

### LINKED ISSUE #140: semantic versioning regression: `0.1.8` followed by `0.1.0`
Debian/buster (currently "stable") ships with `libsamplerate0_0.1.9-2`, which includes a dynamic library file:

    /usr/lib/x86_64-linux-gnu/libsamplerate.so.0.1.8

compiling the latest and greatest libsamplerate-0.2.0 i get a file:

    /usr/lib/x86_64-linux-gnu/libsamplerate.so.0.1.0

This decrements the *micro* release, which is plain wrong!!

the new libsamplerate also introduces a new symbol `src_clone`, which would - according to semantic versioning - warrants a:
- reset of the *micro* version to *0*
- increment of the *minor* version to *2*

please fix this and make a bugfix release asap.

umlaeute:
this attempts to fix #140.

it assumes that the package-version follows [semantic versioning](https://semver.org) for the library itself.
since the `libsamplerate` project *only* contains the library, i think this is a safe assumption.

as libsamplerate is used in high-profile applications, like [vlc](https://www.videolan.org/), [alsa-utils](https://github.com/alsa-project/alsa-utils) or [ardour](https://ardour.org/), it is therefore one of the most popular audio libraries installed on Linux Desktops (e.g. it lists under the [top 520](https://qa.debian.org/popcon.php?package=libsamplerate) most installed packages on an average Debian installation - that popcon lists is probably mainly driven by servers (that don't have much to do with the typical system that needs to do anything with sound) and the like and includes packages  which are essential to actually boot a system - so i think the popcon is a very conservative estimate on the actual popularity of the library on user Desktops).

as such, i think it is really important to get the library versioning right.

Closes: https://github.com/libsndfile/libsamplerate/issues/140

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -73,10 +73,21 @@ dnl      age.
 dnl   6. If any interfaces have been removed since the last public release, then set age
 dnl      to 0.
 
-CLEAN_VERSION=`echo $PACKAGE_VERSION | $SED "s/p.*//"`
-VERSION_MINOR=`echo $CLEAN_VERSION | $SED "s/.*\.//"`
 
-SHARED_VERSION_INFO="1:$VERSION_MINOR:1"
+dnl https://autotools.io/libtool/version.html#idm46482193960416 (2021-01-22)
+dnl For Linux, for instance, while the last two values map directly from the
+dnl command-line, the first is calculated by subtracting age from current.
+dnl On the other hand, on modern FreeBSD, only one component is used in the
+dnl library version, which corresponds to current.
+
+dnl if PACKAGE_VERSION follows [Semantic Versioning](https://semver.org/)
+dnl of the library, we can use this:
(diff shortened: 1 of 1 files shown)


SoapGentoo:
This approach is wrong, because it assumes you can map ABI breaks to API breaks. We'll maintain the `-version` argument ourselves.

umlaeute:
> This approach is wrong, because it assumes you can map ABI breaks to API breaks. 

in the end it depends what the `PACKAGE_VERSION` means *for your project*.

anyhow:

> We'll maintain the -version argument ourselves.

cool.
just don't forget it ;-)

SoapGentoo:
Superseded by #143

PR #143, opened 2021-01-23
Decouple libtool's -version-info from PACKAGE_VERSION

SoapGentoo:
* Trying to compute -version-info from PACKAGE_VERSION is a
  recipe for disaster. We will maintain it separately going
  forward.

Fixes: #140

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,4 +1,6 @@
-Unreleased
+Version 0.2.1 (2021-01-23)
+  * Fix incorrect passing of -version-info to libtool, causing a
+    regression on versioned file name of the shared library (#140).
 
 Version 0.2.0 (2021-01-21)
   * API:
--- a/configure.ac
+++ b/configure.ac
@@ -72,11 +72,7 @@ dnl   5. If any interfaces have been added since the last public release, then i
 dnl      age.
 dnl   6. If any interfaces have been removed since the last public release, then set age
 dnl      to 0.
-
-CLEAN_VERSION=`echo $PACKAGE_VERSION | $SED "s/p.*//"`
-VERSION_MINOR=`echo $CLEAN_VERSION | $SED "s/.*\.//"`
-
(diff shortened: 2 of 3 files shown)


SoapGentoo:
@umlaeute

SoapGentoo:
I now get
```
libsamplerate.so -> libsamplerate.so.0.2.0
libsamplerate.so.0 -> libsamplerate.so.0.2.0
libsamplerate.so.0.2.0
```
which clearly shows that the ABI is backwards compatible, but not forwards compatible.

SoapGentoo:
@umlaeute, I've changed the argument to `2:1:2`, purely to produce the same versioned filename. I do this mostly for cosmetic reasons, so people's OCD doesn't get triggered when seeing a package `0.2.1` producing a library `libsamplerate.so.0.2.0`.

evpobr (approved):

umlaeute:
> @umlaeute, I've changed the argument to 2:1:2, purely to produce the same versioned filename. I do this mostly for cosmetic reasons, so people's OCD doesn't get triggered when seeing a package 0.2.1 producing a library libsamplerate.so.0.2.0.

cool.
so you are basically doing what i suggested in #142 and what you turned down as all wrong. :rofl: 

but anyhow: thanks for getting the version strictly monotonic again, this is really all i care.


sidenote: it seems that you are now maintaining the ABI version in two places `CMakeList.txt` and `configure.ac`.
it would probably be better to just have a single central place, that is used by both build systems.

SoapGentoo:
Yes, because if we have to change an API from `int` -> `long` (for whatever reason), semantic versioning doesn't require a major version bump, and a minor version bump wouldn't change the `SONAME` to indicate an ABI compatibility. By computing this automatically, you're pretending that you can version API and ABI jointly, and gives you a false sense of automated safety.

PR #145, opened 2021-01-23
exclude repository configuration from the archive

umlaeute:
this PR excludes repository configuration (like .gitignore, github-workflows and .travis.yml from the archive files (generated via git archive, or when downloading a zip/tar release (resp. snapshot) from github).

reasoning: those file really only make sense in the context of this gitrepository.

speaking with my Debian maintainer hat on:

if some downstream consumer of the source packages (e.g. a linux distribution package) uses git to manage the packaging, they usually have different requirements on how the repository should be configured, so `.gitignore` files that are perfectly fine for upstream development can often be considered harmful in environments that mostly deal with packaging.

similarily, the `.github/` directory **only** makes sense for a repository that is hosted on github.
`git archive` (and thus: github's "export to zipfile") are obviously not intended to be used to popuplate github repositories (use *fork* for that), so the `.github/` directory shouldn't be included.
simililarily for `.travis.yml`, which is currently not used by libsamplerate

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
+# enforce correct file-extensions
 *.ac text eol=lf
 *.am text eol=lf
 *.m4 text eol=lf
@@ -8,3 +9,7 @@
 *.bat text eol=crlf
 *.def text eol=crlf
 *.msvc text eol=crlf
+
+# repository configuration excluded from the archive
+.git*      	export-ignore
+.travis.yml	export-ignore


SoapGentoo:
I'm fine with this, but doesn't Debian just use the bootstrapped releases? Or do you actually run `autoreconf` on the `git archive` tarballs that Github produces?

umlaeute:
Am 24. Jänner 2021 13:45:17 MEZ schrieb David Seifert <notifications@github.com>:
> I'm fine with this, but doesn't Debian just use the bootstrapped
> releases? Or do you actually run `autoreconf` on the `git archive`
> tarballs that Github produces?


the latter.

Debian uses (at least: might use) patched autotools for better (and uptodate) support on the many supported architectures.
as such it is highly recommended to run `autoreconf` on our own.

the decision to use the auto generated git tarballs rather than bootstrapped release tarballs is at the maintainer's discretion (and I personally very much prefer it, unless of course you - as upstream - don't want that)

SoapGentoo:
Wow, I wasn't aware. In that case, this PR makes much more sense. Gentoo always prefers bootstrapped tarballs, since N users will build the package themselves, and a full `autoreconf` can easily go into 15s for a package that might only take 10s to build.

PR #147, opened 2021-02-10
exported symbols control for macOS dylib builds

sezero:


evpobr:
I'm worried if this will break the ABI?

sezero:
> I'm worried if this will break the ABI?

How's that so?  Without this, the dylib used to export private symbols
along with others.  The exports file takes the list of exported syms from
the windows def file:  If there were to be any ABI breakage, it has been
there for windows and unices already.

evpobr:
Need to be sure. Let's ask @SoapGentoo first.

SoapGentoo:
This has some potential for Hyrum's-Law-ABI breaks. Shouldn't `static` prevent exporting of private symbols? Am I missing something here?

sezero:
> This has some potential for Hyrum's-Law-ABI breaks. Shouldn't `static`
> prevent exporting of private symbols?  Am I missing something here?

I guess you are.  Here is the diff between `nm -gU` outputs from the dylib
built without and with this patch:
```
 
 libsamplerate.dylib(single module):
-000047d0 T _linear_get_description
-000047b0 T _linear_get_name
-000049d0 T _linear_state_new
-00001190 T _sinc_get_description
-00001150 T _sinc_get_name
-000014d0 T _sinc_state_new
 00000ef0 T _src_callback_new
 00000f60 T _src_callback_read
 000006e0 T _src_clone
 00000730 T _src_delete
 000009b0 T _src_error
 00000bd0 T _src_float_to_int_array
 00000c70 T _src_float_to_short_array
 00000920 T _src_get_channels
 00000d10 T _src_get_description
 00000d50 T _src_get_name
 00000990 T _src_get_version
 00000b80 T _src_int_to_float_array
 00001120 T _src_is_valid_ratio
 00000d90 T _src_new
 00000750 T _src_process
 00000940 T _src_reset
 000008d0 T _src_set_ratio
 00000b20 T _src_short_to_float_array
 00000e90 T _src_simple
 000009d0 T _src_strerror
-00004020 T _zoh_get_description
-00004000 T _zoh_get_name
-00004220 T _zoh_state_new
```
Are those private symbols, such as `linear_get_description` static? I guess not.

SoapGentoo:
Yup, but I still have a better idea: in this case, let's use GCC/Clang's symbol visibility instead , which is even [the recommended method by Apple](https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/DynamicLibraries/100-Articles/DynamicLibraryDesignGuidelines.html).

sezero:
> Yup, but I still have a better idea: in this case, let's use GCC/Clang's symbol visibility instead , which is even [the recommended method by Apple]

You can use that not only for macho but for elf, too.

In any case, this patch is not wrong, and I don't have a patch for visibility
attributes.

evpobr:
> Yup, but I still have a better idea: in this case, let's use GCC/Clang's symbol visibility instead , which is even [the recommended method by Apple](https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/DynamicLibraries/100-Articles/DynamicLibraryDesignGuidelines.html).

I like it. But i don't think it is good idea to mix two different methods to export symbols. I vote to use it in next major release.

@SoapGentoo , what if we apply this patch and test it before the next release? If there are any problems, we will revert it.

evpobr:
> I guess you are. Here is the diff between `nm -gU` outputs from the dylib
> built without and with this patch:

@sezero , can we see the output with your patch applied?

sezero:
> @sezero , can we see the output with your patch applied?

Here it is:
```
$ x86_64-apple-darwin9-nm -gU libsamplerate.dylib
0000000000008000 T _linear_get_description
0000000000007fc0 T _linear_get_name
0000000000008040 T _linear_state_new
0000000000001f10 T _sinc_get_description
0000000000001e80 T _sinc_get_name
0000000000001fa0 T _sinc_state_new
0000000000000fd0 T _src_callback_new
0000000000001430 T _src_callback_read
0000000000000f20 T _src_clone
0000000000001140 T _src_delete
0000000000001900 T _src_error
0000000000001da0 T _src_float_to_int_array
0000000000001c60 T _src_float_to_short_array
0000000000001750 T _src_get_channels
0000000000001820 T _src_get_description
0000000000001790 T _src_get_name
00000000000018b0 T _src_get_version
0000000000001d40 T _src_int_to_float_array
00000000000018c0 T _src_is_valid_ratio
0000000000000df0 T _src_new
0000000000001180 T _src_process
00000000000010b0 T _src_reset
00000000000016e0 T _src_set_ratio
0000000000001bf0 T _src_short_to_float_array
0000000000001b70 T _src_simple
0000000000001940 T _src_strerror
00000000000071b0 T _zoh_get_description
0000000000007170 T _zoh_get_name
00000000000071f0 T _zoh_state_new
```

evpobr:
> Here it is:

I see non-static private functions e.g. `_linear_get_description`(). Aren't they supposed to be exported?

configure.ac
```diff
@@ -267,6 +267,8 @@ AS_IF([test "x$ax_cv_c_compiler_vendor" = "xgnu"], [
 		AS_CASE([${host_os}],
 			[mingw*|cygwin*], [
 				SHLIB_VERSION_ARG="-export-symbols \$(top_srcdir)/Win32/libsamplerate-0.def"],
+			[darwin*], [
+				SHLIB_VERSION_ARG="-Wl,-exported_symbols_list,\$(top_srcdir)/src/libsamplerate.exports"],
```
evpobr, line 271: in libsndfile it looks little different:

https://github.com/libsndfile/libsndfile/blob/06fd6f626e75bc57caa9ff43cfa913b66216cd66/configure.ac#L606

sezero:
> > @sezero , can we see the output with your patch applied?
> 
> Here it is:

Sorry, that was built from a wrong branch.  Here it is:
```
$ x86_64-apple-darwin9-nm -gU libsamplerate.dylib
0000000000001500 T _src_callback_new
0000000000001580 T _src_callback_read
0000000000000d70 T _src_clone
0000000000000dc0 T _src_delete
0000000000001000 T _src_error
0000000000001210 T _src_float_to_int_array
00000000000012b0 T _src_float_to_short_array
0000000000000f80 T _src_get_channels
0000000000001350 T _src_get_description
0000000000001390 T _src_get_name
0000000000000ff0 T _src_get_version
00000000000011d0 T _src_int_to_float_array
0000000000001780 T _src_is_valid_ratio
00000000000013d0 T _src_new
0000000000000de0 T _src_process
0000000000000fa0 T _src_reset
0000000000000f40 T _src_set_ratio
0000000000001180 T _src_short_to_float_array
00000000000014a0 T _src_simple
0000000000001010 T _src_strerror
```

configure.ac
sezero: That one has two separate `-Wl`s, mine has two combined into one. Same result.

sezero:
> Sorry, that was built from a wrong branch. Here it is:

No, it _was_ built from the correct branch, but I had used clang in that
particular build.  The restriction `test "x$ax_cv_c_compiler_vendor" = "xgnu"`
was in the way: You _really_ should enable clang there.

sezero:
> > Sorry, that was built from a wrong branch. Here it is:
> 
> No, it _was_ built from the correct branch, but I had used clang in that
> particular build. The restriction `test "x$ax_cv_c_compiler_vendor" = "xgnu"`
> was in the way: You _really_ should enable clang there.

Just did that with 2de34be

sezero:
All seems to work now after  b4f65b0.

evpobr (changes requested):
Squash commits please.

sezero:
> Squash commits please.

Done.

SoapGentoo:
Given the number of private functions we want to hide, I feel this solution is way out of proportion. The visibility decorators are easy and can be guarded with simple macro guards, and don't require complex build system hackery, hence I'm opposed to the current solution as is. I'll bring up a PR on the weekend with my alternative suggestion.

SoapGentoo (changes requested):
See #148 for a simpler and more scalable approach

SoapGentoo:
#148 has been merged

PR #148, opened 2021-02-11
Hide more private symbols using GCC's visibility support

SoapGentoo:
@evpobr @sezero 
My diff, which I tested on macOS:

```
--- before
+++ after
@@ -1,9 +1,3 @@
-0000000000008d00 T _linear_get_description
-0000000000008ce0 T _linear_get_name
-0000000000008d20 T _linear_state_new
-0000000000004eb0 T _sinc_get_description
-0000000000004e90 T _sinc_get_name
-0000000000004ed0 T _sinc_state_new
 0000000000004140 T _src_callback_new
 00000000000044b0 T _src_callback_read
 00000000000040f0 T _src_clone
@@ -24,6 +18,3 @@
 0000000000004b10 T _src_short_to_float_array
 00000000000048f0 T _src_simple
 00000000000048d0 T _src_strerror
-0000000000007a90 T _zoh_get_description
-0000000000007a70 T _zoh_get_name
-0000000000007ab0 T _zoh_state_new
```

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -1,5 +1,5 @@
 /*
-** Copyright (c) 2002-2016, Erik de Castro Lopo <erikd@mega-nerd.com>
+** Copyright (c) 2002-2021, Erik de Castro Lopo <erikd@mega-nerd.com>
 ** All rights reserved.
 **
 ** This code is released under 2-clause BSD license. Please see the
@@ -15,6 +15,18 @@
 #endif
 
 #include <math.h>
+
+#if defined (_WIN32) || defined (__CYGWIN__) || defined (__MORPHOS__)
+  #define LIBSAMPLERATE_DLL_PRIVATE
+#else
+  #if __GNUC__ >= 4 /* also works for Clang */
+    #define LIBSAMPLERATE_DLL_PRIVATE __attribute__ ((visibility ("hidden")))
+  #elif defined (__APPLE__)
+    #define LIBSAMPLERATE_DLL_PRIVATE __private_extern__
+  #else
+    #define LIBSAMPLERATE_DLL_PRIVATE
+  #endif
+#endif
 
 #define	SRC_MAX_RATIO			256
 #define	SRC_MAX_RATIO_STR		"256"
--- a/src/src_linear.c
+++ b/src/src_linear.c
@@ -1,5 +1,5 @@
 /*
-** Copyright (c) 2002-2016, Erik de Castro Lopo <erikd@mega-nerd.com>
+** Copyright (c) 2002-2021, Erik de Castro Lopo <erikd@mega-nerd.com>
 ** All rights reserved.
 **
 ** This code is released under 2-clause BSD license. Please see the
@@ -155,7 +155,7 @@ linear_vari_process (SRC_STATE *state, SRC_DATA *data)
 /*------------------------------------------------------------------------------
 */
 
-const char*
+LIBSAMPLERATE_DLL_PRIVATE const char*
 linear_get_name (int src_enum)
 {
 	if (src_enum == SRC_LINEAR)
@@ -164,7 +164,7 @@ linear_get_name (int src_enum)
 	return NULL ;
 } /* linear_get_name */
 
-const char*
+LIBSAMPLERATE_DLL_PRIVATE const char*
 linear_get_description (int src_enum)
 {
 	if (src_enum == SRC_LINEAR)
@@ -193,7 +193,7 @@ linear_data_new (int channels)
 	return priv ;
 }
 
-SRC_STATE *
+LIBSAMPLERATE_DLL_PRIVATE SRC_STATE *
 linear_state_new (int channels, SRC_ERROR *error)
 {
 	assert (channels > 0) ;
--- a/src/src_sinc.c
+++ b/src/src_sinc.c
@@ -1,5 +1,5 @@
 /*
-** Copyright (c) 2002-2016, Erik de Castro Lopo <erikd@mega-nerd.com>
+** Copyright (c) 2002-2021, Erik de Castro Lopo <erikd@mega-nerd.com>
 ** All rights reserved.
 **
 ** This code is released under 2-clause BSD license. Please see the
@@ -164,7 +164,7 @@ int_div_ceil (int divident, int divisor) /* == (int) ceil ((float) divident / di
 /*----------------------------------------------------------------------------------------
 */
 
-const char*
+LIBSAMPLERATE_DLL_PRIVATE const char*
 sinc_get_name (int src_enum)
 {
 	switch (src_enum)
@@ -183,7 +183,7 @@ sinc_get_name (int src_enum)
 	return NULL ;
 } /* sinc_get_descrition */
 
-const char*
+LIBSAMPLERATE_DLL_PRIVATE const char*
 sinc_get_description (int src_enum)
 {
 	switch (src_enum)
@@ -257,7 +257,7 @@ sinc_filter_new (int converter_type, int channels)
 	return priv ;
 }
 
-SRC_STATE *
+LIBSAMPLERATE_DLL_PRIVATE SRC_STATE *
 sinc_state_new (int converter_type, int channels, SRC_ERROR *error)
 {
 	assert (converter_type == SRC_SINC_FASTEST ||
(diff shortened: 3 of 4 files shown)


src/common.h
```diff
@@ -16,6 +16,12 @@
 
 #include <math.h>
 
+#if __GNUC__ >= 4 /* also works for Clang */
```
sezero, line 19: You must check for __GNUC__ being defined and also exclude windows here.
Something like:
`#if !defined(_WIN32) && (defined(__GNUC__) && __GNUC__ >= 4) /* also works for Clang */`

SoapGentoo: in general, you dont need to check if a macro is defined before using it. if `__GNUC__` isn't defined, the clause will be false. The Win32 stuff is correct though.

src/common.h
```diff
@@ -16,6 +16,16 @@
...
 
+#if defined _WIN32 || defined __CYGWIN__
+  #define LIBSAMPLERATE_DLL_PRIVATE
+#else
+  #if __GNUC__ >= 4 /* also works for Clang */
+    #define LIBSAMPLERATE_DLL_PRIVATE __attribute__ ((visibility ("hidden")))
+  #else
+    #define LIBSAMPLERATE_DLL_PRIVATE
+  #endif
+#endif
```
SoapGentoo, line 27: taken from https://gcc.gnu.org/wiki/Visibility

sezero: This looks good to me.

Maybe add the following before the #else case? (completely optional,
for ancient macs only)
```
  #elif defined(__APPLE__)
    #define LIBSAMPLERATE_DLL_PRIVATE __private_extern__
```

sezero: One less important note:, some other exotic platforms might not support visibility attributes.
E.g. tried building for morphos, I got:
```
../src/src_sinc.c: In function ‘sinc_get_name’:
../src/src_sinc.c:184: warning: visibility attribute not supported in this configuration; ignored
../src/src_sinc.c: In function ‘sinc_get_description’:
../src/src_sinc.c:204: warning: visibility attribute not supported in this configuration; ignored
../src/src_sinc.c: In function ‘sinc_state_new’:
../src/src_sinc.c:309: warning: visibility attribute not supported in this configuration; ignored
../src/src_zoh.c: In function ‘zoh_get_name’:
../src/src_zoh.c:154: warning: visibility attribute not supported in this configuration; ignored
../src/src_zoh.c: In function ‘zoh_get_description’:
../src/src_zoh.c:163: warning: visibility attribute not supported in this configuration; ignored
../src/src_zoh.c: In function ‘zoh_state_new’:
../src/src_zoh.c:216: warning: visibility attribute not supported in this configuration; ignored
../src/src_linear.c: In function ‘linear_get_name’:
../src/src_linear.c:165: warning: visibility attribute not supported in this configuration; ignored
../src/src_linear.c: In function ‘linear_get_description’:
../src/src_linear.c:174: warning: visibility attribute not supported in this configuration; ignored
../src/src_linear.c: In function ‘linear_state_new’:
../src/src_linear.c:227: warning: visibility attribute not supported in this configuration; ignored
```
Maybe add a check to configury?

SoapGentoo: https://clang.llvm.org/docs/DiagnosticsReference.html#wprivate-extern
Apparently modern clang warns about `__private_extern__`? Not sure this is the right tradeoff.
how would we test the visibility attribute for MorphOS? In general, because it only produces a warning, a compile test won't bomb out, and hence won't easily be detectable. Ideas? `-Werror`?

SoapGentoo: I found https://www.gnu.org/software/autoconf-archive/ax_gcc_func_attribute.html this. Could you give it a try on MorphOS? If it works, we can add it.

sezero: > Apparently modern clang warns about __private_extern__? Not sure this is the right tradeoff.

Of course it will. But you won't hit that with clang which reports itself as gcc4.2, i.e. the top level gcc>4 check applies.

sezero: > I found https://www.gnu.org/software/autoconf-archive/ax_gcc_func_attribute.html this.
> Could you give it a try on MorphOS? If it works, we can add it.

Will look at it.

sezero: > I found https://www.gnu.org/software/autoconf-archive/ax_gcc_func_attribute.html this.
> Could you give it a try on MorphOS? If it works, we can add it.

I added `AX_GCC_FUNC_ATTRIBUTE(visibility)` to configure.ac, with `-Werror`
added to CFLAGS, result is:  `checking for __attribute__((visibility))... yes`
Apparently doesn't work.

SoapGentoo: looking at the m4, it checks whether there's an `-Wattributes` warning, which likely only catches GCC warnings, so this macro is mostly useless.
I think the only good solution here is to have platform-specific blacklists. https://sourceforge.net/p/predef/wiki/OperatingSystems/ tells me that `__MORPHOS__` should work?

SoapGentoo: @sezero added `__private_extern__` and `__MORPHOS__`, please give it another try

evpobr: https://fossies.org/linux/gdb/gnulib/import/m4/visibility.m4

sezero: > @sezero added `__private_extern__` and `__MORPHOS__`, please give it another try

The `__private_extern__` is in the wrong place. It should look like:
```
+#if defined _WIN32 || defined __CYGWIN__ || defined __MORPHOS__
+  #define LIBSAMPLERATE_DLL_PRIVATE
+#else
+  #if __GNUC__ >= 4 /* also works for Clang */
+    #define LIBSAMPLERATE_DLL_PRIVATE __attribute__ ((visibility ("hidden")))
+  #elif defined(__APPLE__)
+    #define LIBSAMPLERATE_DLL_PRIVATE __private_extern__
+  #else
+    #define LIBSAMPLERATE_DLL_PRIVATE
+  #endif
+#endif
```

MorphOS is OK.

SoapGentoo: @evpobr do we really want to pull in gnulib? I'm not totally opposed, but people will curse, because it's literally the kitchen sink.

evpobr: If you're both happy, I don't mind. But I want full confidence that nothing will break.

I'm saying this, I've already broken the libsndfile ABI in much the same way:

libsndfile/libsndfile#268

As it turned out, at least the export script for Linux contained more important information than just a list of public functions. And that broke ABI.

SoapGentoo: @sezero thanks, messed that up, should be fixed now.
@evpobr fortunately, we're not removing the `-Wl,--version-script` argument here. That said, if some consumer pasted `SRC_STATE* linear_state_new (int channels, SRC_ERROR *error)` into their codebase, this change will break it, because these private-but-publicly-linkable symbols are now genuinely private and you'd get a linker error. Given that they were never part of the public interface, I don't generally have a problem with breaking people relying on our implementation details.

sezero: > @evpobr do we really want to pull in gnulib? I'm not totally opposed, but people will curse, because it's literally the kitchen sink.

We don't have to pull in gnulib, we just add visibility.m4 under m4/
but only after modifying it like the following:
```
--- visibility.m4~
+++ visibility.m4
@@ -58,6 +58,7 @@
               extern __attribute__((__visibility__("default"))) int exportedvar;
               extern __attribute__((__visibility__("hidden"))) int hiddenfunc (void);
               extern __attribute__((__visibility__("default"))) int exportedfunc (void);
+              void dummyfunc (void);
               void dummyfunc (void) {}
             ]],
             [[]])],
```
Then it successfully rejects morphos:
```
cc1: warnings being treated as errors
conftest.c: In function 'dummyfunc':
conftest.c:54: error: visibility attribute not supported in this configuration; ignored
configure:15498: $? = 1
```
And it defines `HAVE_VISIBILITY` either to 0 or 1 in config.h

SoapGentoo: which means you have to get it upstreamed, because someone, somewhere down the road will pull a fresh copy and replace that again. Imo, not worth the hassle.

SoapGentoo: I just checked the ELF section headers, they're all still there. So yes, I'm just as confident that this won't break anything, modulo consumers relying on implementation details.

sezero: > which means you have to get it upstreamed, because someone, somewhere down the road will pull a fresh copy and replace that again. Imo, not worth the hassle.

FWIW, just sent a patch, here: https://lists.gnu.org/archive/html/bug-gnulib/2021-02/msg00052.html

evpobr (approved):

src/common.h
SoapGentoo: @sezero can we merge this now?

sezero: Sure.

evpobr:
@SoapGentoo , it is your PR.

Issue #149, opened 2021-02-11
condition for using Version_script and the win32 def file


sezero:
Our current configure.ac checks gcc (and only gcc) being the compiler
when deciding whether to use Version_script or the win32 def file:
https://github.com/libsndfile/libsamplerate/blob/master/configure.ac#L265

However, the linker should be one checked. Do you think the following
is correct: @SoapGentoo, @evpobr?  If it is I can make a PR for it.
```
diff --git a/configure.ac b/configure.ac
index 0e2117c..4691e09 100644
--- a/configure.ac
+++ b/configure.ac
@@ -260,14 +260,13 @@ AS_IF([test "x$enable_werror" = "xyes"], [
 AX_APPEND_COMPILE_FLAGS([-W -Wstrict-prototypes -Wmissing-prototypes -Wall -Waggregate-return -Wcast-align -Wcast-qual -Wnested-externs -Wshadow -Wpointer-arith], [CFLAGS])
 
 dnl ====================================================================================
-dnl  GCC stuff.
+dnl  Exported symbols control.
 
-AS_IF([test "x$ax_cv_c_compiler_vendor" = "xgnu"], [
-		# OS specific tweaks.
+AS_IF([test "x$lt_cv_prog_gnu_ld" = "xyes"], [
 		AS_CASE([${host_os}],
 			[mingw*|cygwin*], [
 				SHLIB_VERSION_ARG="-export-symbols \$(top_srcdir)/Win32/libsamplerate-0.def"],
-			[linux*|kfreebsd*-gnu*|gnu*], [
+			[*], [
 				SHLIB_VERSION_ARG="-Wl,--version-script=\$(top_builddir)/src/Version_script"
 			])
 	])
```

evpobr:
If Clang is not detected by this path it makes sense.

sezero:
> If Clang is not detected by this path 

It certainly is not detected (had noticed that in #147)

@SoapGentoo?

SoapGentoo:
This seems like the right idea. I'd consider going even further, dumping the whole `AS_CASE` around it and getting rid of the `-export-symbols` cruft in favour of the `LIBSAMPLERATE_DLL_*` approach. @sezero what do you think? You'd end up with just
```
AS_IF([test "x$lt_cv_prog_gnu_ld" = "xyes"], [
		SHLIB_VERSION_ARG="-Wl,--version-script=\$(top_builddir)/src/Version_script"
	])
```
[...]

PR #150, opened 2021-02-12
Check GNU linker instead of gcc compiler for exported symbols control.

sezero:
Also enable unix OSs other than linux, kfreebsd and hurd.  macos isn't
affected because it doesn't use GNU ld.

CMake would need something similar, i.e. compiler id -> linker id check,
but I don't know cmake enough to do that.

Closes: #149

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,7 @@
 Unreleased
   * Switch to GCC's visibility for hiding more implementation details
+  * Check GNU ld instead of gcc for exported symbols control logic in
+    configure.ac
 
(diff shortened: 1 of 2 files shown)


SoapGentoo:
all fine, could you just add a ntoe to `NEWS` something along the lines of `Fixing GNU ld detection logic in configure.ac`

sezero:
> all fine, could you just add a ntoe to `NEWS` something along the lines of `Fixing GNU ld detection logic in configure.ac`

Done

PR #151, opened 2021-02-13
check visibility attributes support in configury and cmake

sezero:
This adds checks for visibility attributes support to configury and cmake.
CMake side is a simple addition of a few lines.  The configury (autotools)
side relies on visibility.m4 taken from gnulib, which recently got a fix
for -Wmissing-prototypes:
http://git.savannah.gnu.org/gitweb/?p=gnulib.git;a=commit;h=802931d3728acb52763c13e2644c06ad2ce0cc3f

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -10,11 +10,20 @@ set(ENABLE_SINC_FAST_CONVERTER ${LIBSAMPLERATE_ENABLE_SINC_FAST_CONVERTER} PAREN
 set(ENABLE_SINC_MEDIUM_CONVERTER ${LIBSAMPLERATE_ENABLE_SINC_MEDIUM_CONVERTER} PARENT_SCOPE)
 set(ENABLE_SINC_BEST_CONVERTER ${LIBSAMPLERATE_ENABLE_SINC_BEST_CONVERTER} PARENT_SCOPE)
 
+include(CMakePushCheckState)
+include(CheckCSourceCompiles)
 include(ClipMode)
 include(CMakePackageConfigHelpers)
 
 # This will set CPU_CLIPS_NEGATIVE and CPU_CLIPS_POSITIVE
 clip_mode()
+
+# check for symbol visibility attributes support
+cmake_push_check_state()
+set(CMAKE_REQUIRED_FLAGS "-fvisibility=hidden -Werror")
+check_c_source_compiles("int foo(void) __attribute__((visibility(\"default\")));
+                         int main(void) {return 0;}" HAVE_VISIBILITY)
+cmake_pop_check_state()
 
 add_library(samplerate
   common.h
--- a/config.h.cmake
+++ b/config.h.cmake
@@ -54,6 +54,10 @@
 /* Define to 1 if you have the <unistd.h> header file. */
 #cmakedefine HAVE_UNISTD_H
 
+/* Define to 1 or 0, depending whether the compiler supports simple visibility
+   declarations. */
+#cmakedefine01 HAVE_VISIBILITY
+
 /* define fast samplerate convertor */
 #cmakedefine ENABLE_SINC_FAST_CONVERTER
 
--- a/configure.ac
+++ b/configure.ac
@@ -262,6 +262,8 @@ AX_APPEND_COMPILE_FLAGS([-W -Wstrict-prototypes -Wmissing-prototypes -Wall -Wagg
 dnl ====================================================================================
 dnl  Exported symbols control.
 
+gl_VISIBILITY
+
 AS_IF([test "x$lt_cv_prog_gnu_ld" = "xyes"], [
 		AS_CASE([${host_os}],
 			[mingw*|cygwin*], [
--- a/m4/visibility.m4
+++ b/m4/visibility.m4
@@ -1,0 +1,78 @@
+# visibility.m4 serial 7
+dnl Copyright (C) 2005, 2008, 2010-2021 Free Software Foundation, Inc.
+dnl This file is free software; the Free Software Foundation
+dnl gives unlimited permission to copy and/or distribute it,
+dnl with or without modifications, as long as this notice is preserved.
+
+dnl From Bruno Haible.
+
+dnl Tests whether the compiler supports the command-line option
+dnl -fvisibility=hidden and the function and variable attributes
+dnl __attribute__((__visibility__("hidden"))) and
+dnl __attribute__((__visibility__("default"))).
+dnl Does *not* test for __visibility__("protected") - which has tricky
+dnl semantics (see the 'vismain' test in glibc) and does not exist e.g. on
+dnl Mac OS X.
+dnl Does *not* test for __visibility__("internal") - which has processor
+dnl dependent semantics.
+dnl Does *not* test for #pragma GCC visibility push(hidden) - which is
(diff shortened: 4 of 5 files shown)


src/CMakeLists.txt
```diff
@@ -10,12 +10,20 @@ set(ENABLE_SINC_FAST_CONVERTER ${LIBSAMPLERATE_ENABLE_SINC_FAST_CONVERTER} PAREN
...
 
 # This will set CPU_CLIPS_NEGATIVE and CPU_CLIPS_POSITIVE
 clip_mode()
 
+# check for symbol visibility attributes support
+set(OLD_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS}")
+set(CMAKE_REQUIRED_FLAGS "-fvisibility=hidden -Werror ${OLD_REQUIRED_FLAGS}")
+check_c_source_compiles("int foo(void) __attribute__((visibility(\"default\")));
+                         int main(void) {return 0;}" HAVE_VISIBILITY)
+set(CMAKE_REQUIRED_FLAGS "${OLD_REQUIRED_FLAGS}")
```
evpobr, line 25: I would use [CMakePushCheckState](https://cmake.org/cmake/help/latest/module/CMakePushCheckState.html).

evpobr (changes requested):

src/CMakeLists.txt
sezero: Per your suggestion I changed it to this:
```
cmake_push_check_state()
set(CMAKE_REQUIRED_FLAGS "-fvisibility=hidden -Werror")
check_c_source_compiles("int foo(void) __attribute__((visibility(\"default\")));
                         int main(void) {return 0;}" HAVE_VISIBILITY)
cmake_pop_check_state()
```

It seemed to work, _however_ when I used a toolchain file, like
`-DCMAKE_TOOLCHAIN_FILE=Toolchain-MinGW64.cmake` it failed:
```
-- Checking processor clipping capabilities...
-- Checking processor clipping capabilities... disabled
CMake Error at src/CMakeLists.txt:21 (cmake_push_check_state):
  Unknown CMake command "cmake_push_check_state".


-- Configuring incomplete, errors occurred!
```

This is with cmake 3.9.6.  If it matters, the toolchain file looks like this:
```
# toolchain file I use to cross compile on Linux
# targetting Windows (x64/mingw-w64). running:
# cmake -DCMAKE_TOOLCHAIN_FILE=/path/to/Toolchain-MinGW64.cmake ....

SET(CMAKE_SYSTEM_NAME Windows)

SET(CMAKE_C_COMPILER /opt/cross_win64/bin/x86_64-w64-mingw32-gcc)
SET(CMAKE_CXX_COMPILER /opt/cross_win64/bin/x86_64-w64-mingw32-g++)
SET(CMAKE_RC_COMPILER /opt/cross_win64/bin/x86_64-w64-mingw32-windres)

# where is the target environment
SET(CMAKE_FIND_ROOT_PATH /opt/cross_win64 /opt/cross_win64/x86_64-w64-mingw32)

# adjust the default behaviour of the FIND_XXX() commands:
# search headers and libraries in the target environment, search
# programs in the host environment
SET(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
SET(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
SET(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
```

SoapGentoo: @sezero what's the gain over the simple "whitelist" approach we currently have? I'm not opposed, but we can cover 98% of users with the current macro cases and don't have to pull in a bunch of cmake and autoconf macros.

sezero: > @sezero what's the gain over the simple "whitelist" approach we currently have? I'm not opposed, but we can cover 98% of users with the current macro cases and don't have to pull in a bunch of cmake and autoconf macros.

IMO, a whitelist/blacklist approach should be against a well-known well-tested common case provided by autoconf or cmake.

evpobr: Why not see how the CMake authors did it: https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/GenerateExportHeader.cmake

`_test_compiler_hidden_visibility` macro looks interesting. In fact, I strongly suspect that we might be using a `COMPILER_HAS_HIDDEN_VISIBILITY` variable.

sezero: I found what I did wrong with CMakePushCheckState: I had missed a cmake include, now it works.

As for how the CMake authors did it: I don't see how my three lines are inferior or wrong?

Force-pushed my last version.

evpobr: There are some corner cases, but they are rare. We can use current version.

sezero: > There are some corner cases, but they are rare. We can use current version.

OK, anything else needed from me?

sezero:
Unless anything else needed from me, this is my final version of this patch.

evpobr (approved):

evpobr (approved):

src/CMakeLists.txt
```diff
@@ -10,12 +10,21 @@ set(ENABLE_SINC_FAST_CONVERTER ${LIBSAMPLERATE_ENABLE_SINC_FAST_CONVERTER} PAREN
...
+include(CheckCSourceCompiles)
 include(ClipMode)
 include(CMakePackageConfigHelpers)
 
 # This will set CPU_CLIPS_NEGATIVE and CPU_CLIPS_POSITIVE
 clip_mode()
 
+# check for symbol visibility attributes support
+cmake_push_check_state()
+set(CMAKE_REQUIRED_FLAGS "-fvisibility=hidden -Werror")
```
SoapGentoo, line 23: ...does this actually work on MSVC?

SoapGentoo (changes requested):

src/CMakeLists.txt
sezero: > ...does this actually work on MSVC?

Of course not. I just avoided multiple if()/endif()s.

SoapGentoo: how do we get this to work for `__declspec(dllexport)` at a later stage then? add another `#elif defined (_WIN32) || defined (__CYGWIN__)` macro clause?

sezero: > how do we get this to work for `__declspec(dllexport)` at a later stage then? add another `#elif defined (_WIN32) || defined (__CYGWIN__)` macro clause?

We don't. src/CMakeLists.txt uses the def file for exports:
https://github.com/libsndfile/libsamplerate/blob/master/src/CMakeLists.txt#L19

SoapGentoo: oh right, sorry, forgot that already :rofl:

sezero:
Are there any more changes requested or anything else left to do with this one?

SoapGentoo:
I'll test it on the weekend. I'm ultimately still not super amped about adding these macros over just maintaining a white list.

SoapGentoo:
All fine, just fix the typo in the commitmsg:
`check visibility attributes support in configure and cmake`

sezero:
> All fine, just fix the typo in the commitmsg:
> `check visibility attributes support in configure and cmake`

Done (it wasn't a typo though :)

PR #152, opened 2021-02-16
libm is not needed for mingw / darwin

sezero:


evpobr:
Yes, we have problem here.

But here are my thoughts:

CMake:

```cmake
check_function_exists(sin HAVE_DEFAULT_MATH)
if(NOT HAVE_DEFAULT_MATH)
  check_library_exists(m sin "" HAVE_LIBM)
endif()
```

For Autotools: [`LT_LIB_M` macro](https://www.gnu.org/software/libtool/manual/html_node/Autoconf-macros.html).

sezero:
> Yes, we have problem here.
> 
> But here are my thoughts:
> 
> CMake:
> 
> ```cmake
> check_function_exists(sin HAVE_DEFAULT_MATH)
> if(NOT HAVE_DEFAULT_MATH)
>   check_library_exists(m sin "" HAVE_LIBM)
> endif()
> ```

OK, will change things

> For Autotools: [`LT_LIB_M` macro](https://www.gnu.org/software/libtool/manual/html_node/Autoconf-macros.html).

Looked at it, `LT_LIB_M` exclude many to not need libm but not mingw,
possibly another libtool oversights among many.  I can duplicate it. *Or*,
for better, I can do `AC_CHECK_FUNC(floor, need_libm=no, need_libm=yes)`
just like your cmake suggestion.

sezero:
> > For Autotools: [`LT_LIB_M` macro](https://www.gnu.org/software/libtool/manual/html_node/Autoconf-macros.html).
> 
> Looked at it, `LT_LIB_M` exclude many to not need libm but not mingw,

FWIW, sent a patch to mainstream libtool for it, here: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=46559

sezero:
Updated the patch: Used filter lists based on LT_LIB_M of libtool-2.4.6
(with mingw added),  and with cmake side made to mirror it as much as
possible.

I found that the `check_function_exists(sin HAVE_DEFAULT_MATH)`
approach (or its autotools equivalents) might result in false positives,
where some C89 math functions are in libc but some C99 ones (e.g. 
the ones needed here) reside in libm:  So I'm preferring the filter list
approach.

evpobr:
Ok. I can merge it?

sezero:
> Ok. I can merge it?

I have nothing else to add to it.

evpobr:
Thanks!

PR #157, opened 2021-05-23
Test ac update

SoapGentoo:


SoapGentoo:
@evpobr cleaning up before the ABI fix

evpobr:
Why disable static build?

SoapGentoo:
_by default_, you can't disable static libraries in libtool. The point is, don't produce static libraries unless the user asks for them

configure.ac
```diff
@@ -54,7 +54,7 @@ dnl Catch ancient versions of pkg-config below 0.27
 AX_REQUIRE_DEFINED([PKG_INSTALLDIR])
 PKG_INSTALLDIR
 
-LT_INIT
+LT_INIT([disable-static])
```
evpobr, line 57: I guess it is better to add info about new behavior to CHANGELOG.md.

evpobr (changes requested):

configure.ac
SoapGentoo: better now?

evpobr (approved):

PR #158, opened 2021-05-24
Fix macos abi

SoapGentoo:
@evpobr 
While copying your pattern from libsndfile, I think you've made a mistake in emulating [libtool's behaviour](https://git.savannah.gnu.org/cgit/libtool.git/tree/build-aux/ltmain.in#n6997):

    -compatibility_version $(current + 1) -current_version $(current + 1).$revision

on my Mac, I see it passing `-compatibility_version 3 -current_version 3.1` in Automake.

With this branch, CMake and Autotools should be aligned again:
```
$ otool -L cmake-master/lib/libsamplerate.dylib
cmake-master/lib/libsamplerate.dylib:
	@rpath/libsamplerate.0.dylib (compatibility version 0.0.0, current version 0.2.1)

$ otool -L autotools-master/lib/libsamplerate.dylib
autotools-master/lib/libsamplerate.dylib:
	[...]/autotools-master/lib/libsamplerate.0.dylib (compatibility version 3.0.0, current version 3.1.0)

$ otool -L cmake-branch/lib/libsamplerate.dylib
cmake-branch/lib/libsamplerate.dylib:
	@rpath/libsamplerate.0.dylib (compatibility version 3.0.0, current version 3.1.0)
```


evpobr:
> While copying your pattern from libsndfile, I think you've made a mistake in emulating libtool's behaviour:

Where is the bug?

```cmake
set (SNDFILE_ABI_VERSION_MAJOR 0)
set (SNDFILE_ABI_VERSION_MINOR 2)
set (SNDFILE_ABI_VERSION_PATCH 1)

# 0.2.1
set (SNDFILE_ABI_VERSION "${SNDFILE_ABI_VERSION_MAJOR}.${SNDFILE_ABI_VERSION_MINOR}.${SNDFILE_ABI_VERSION_PATCH}")

# 0 + 2 + 1 = 3
math (EXPR SNDFILE_MACHO_CURRENT_VERSION_MAJOR "${SNDFILE_ABI_VERSION_MAJOR} + ${SNDFILE_ABI_VERSION_MINOR} + 1")

# current version 3.1.0
set (SNDFILE_MACHO_CURRENT_VERSION "${SNDFILE_MACHO_CURRENT_VERSION_MAJOR}.${SNDFILE_ABI_VERSION_PATCH}.0")

# ompatibility version 3.0.0
set (SNDFILE_MACHO_COMPATIBILITY_VERSION "${SNDFILE_MACHO_CURRENT_VERSION_MAJOR}.0.0")
```

I think you try to calculate CMake ABI versions here from libtool version, not CMake's ABI version format.

evpobr:
Actually i have idea how to fix that mess - CMake project must read and extract via regex `lt_current`, `lt_revision` and `lt_age` values from `configure.ac` and calculate ABI versions for CMake.

evpobr (approved):
You code is valid anyway.

SoapGentoo:
I think that's better than creating even more indirection. That said, here's my suggestion, let's just bind ABI to API version, this is getting too complicated. If we need to break the ABI, we'll bump the major version too?

evpobr:
> I think that's better than creating even more indirection. That said, here's my suggestion, let's just bind ABI to API version, this is getting too complicated. If we need to break the ABI, we'll bump the major version too?

Don't worry, this was a quick fix for macOS compatibility. Before new release I will make a patch, after which you will only need to bump the libtool version in `configure.ac`.

PR #160, opened 2021-05-27
cmake, autotools: better check for visibility attributes support.

sezero:
in autotools case, also moves warning flags additions after visibility
checks so that false negatives wouldn't happen and removes dependence on
gnulib's visibility.m4.

Fixes: https://github.com/libsndfile/libsamplerate/issues/154


evpobr:
What's wrong with visibility? It was fixed by 70423ce.

sezero:
> What's wrong with visibility? It was fixed by [70423ce](https://github.com/libsndfile/libsamplerate/commit/70423ce8c77b8cb32c5ee18dd30e866ebafd22f3).

It was a band-aid, this should be a more general solution.

evpobr (approved):

SoapGentoo (approved):

Issue #159, opened 2021-05-27
CMake Error at CMakeLists.txt:103 (add_library)


sezero:
```
-- Checking for module 'sndfile'
--   Found sndfile, version 1.0.20
CMake Error at CMakeLists.txt:103 (add_library):
  add_library cannot create ALIAS target "SndFile::sndfile" because target
  "PkgConfig::SndFile" is IMPORTED.
```
This is with cmake 3.9.6 and 3.11.2

evpobr:
@sezero , check please #161.

sezero:
> @sezero , check please #161.

Works for me - tested with cmake-3.9.6.

PR #161, opened 2021-05-27
Fix dependency seach for CMake <= 3.13

evpobr:


sezero:
This does fix #159 for me - tested with cmake 3.9.6 and 3.11.2.
The only issue is that with cmake-3.11.2, I get a _weird_ message
when libsndfile-devel and/or fftw3-devel is not installed:
```
-- Checking for module 'sndfile'
--   Package 'sndfile', required by 'virtual:world', not found
[...]
-- Checking for module 'fftw3'
--   Package 'fftw3', required by 'virtual:world', not found
```
What's with `virtual:world`?.. But that's a non-issue.

evpobr:
> This does fix #159 for me - tested with cmake 3.9.6 and 3.11.2.
> The only issue is that with cmake-3.11.2, I get a _weird_ message
> when libsndfile-devel and/or fftw3-devel is not installed:
> 
> ```
> -- Checking for module 'sndfile'
> --   Package 'sndfile', required by 'virtual:world', not found
> [...]
> -- Checking for module 'fftw3'
> --   Package 'fftw3', required by 'virtual:world', not found
> ```
> 
> What's with `virtual:world`?.. But that's a non-issue.

Hmm, i don't know.

PR #177, opened 2022-05-30
fix BUILD_TESTING option

ferluht:
BUILD_TESTING was always set to ON by include(CTest)
Now LIBSAMPLERATE_TESTS can be used to disable tests build

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,11 +42,12 @@ include(CheckLibraryExists)
 include(CheckSymbolExists)
 include(GNUInstallDirs)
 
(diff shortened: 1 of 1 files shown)


evpobr:
Hi @ferluht . To disable tests set `BUILD_TESTING` to `OFF` from command line, CTest module will not enable it again if it disabled manually. Documentation lack here.

```
cmake .. -DBUILD_TESTING=OFF
```

evpobr (approved):

evpobr:
thanks @ferluht .

PR #178, opened 2022-06-16
cmake: limit dll name to 8 characters for os/2.

sezero:


evpobr:
Hi @sezero !

src/CMakeLists.txt
```diff
@@ -50,12 +50,17 @@ if(BUILD_SHARED_LIBS AND WIN32)
...
-  endif()
+    endif()
   else()
     if(MINGW OR CYGWIN)
       set_target_properties(samplerate PROPERTIES RUNTIME_OUTPUT_NAME "samplerate-${libsamplerate_VERSION_MAJOR}")
     endif()
   endif()
+  if(OS2)
+    # OS/2 doesn't support a DLL name longer than 8 characters.
+    set_target_properties(samplerate PROPERTIES OUTPUT_NAME "SAMPRATE")
```
evpobr, line 61: Should it be `RUNTIME_OUTPUT_NAME`?

sezero: I'm not much of a cmake guy:  if the effect is the same, i.e. the generated dll will be named SAMPRATE.dll, then I'll make the change and force-push.

evpobr: Ok, let's  leave as it is. I don't know how it is more correct here, this is not a very popular operating system.

evpobr:
Thanks for help!

PR #188, opened 2023-01-13
Implement SSE2 lrint() and lrintf()

### LINKED ISSUE #187: Slow `lrint()` `lrintf()` functions in the release Windows build.
The `SRC_LINEAR` method is about 10 times slower in the default MSVC build compared to the MinGW64 build. MinGW64 uses custom `lrint()` [implementation](https://github.com/mingw-w64/mingw-w64/blob/4359b3570a717d20dcb4251f5b5e94459e1a64ac/mingw-w64-headers/crt/math.h#L977-L983). To solve this problem, I suggest adding [/fp:fast](https://learn.microsoft.com/en-us/cpp/build/reference/fp-specify-floating-point-behavior?view=msvc-170#fast) option:
```
if(MSVC)
  add_compile_options("/fp:fast")
endif()
```
An alternative solution is probably to resurrect the `float_cast.h` file for the MSVC case?

rfomin:
Fix #187

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,6 +88,7 @@ endif()
 
 check_include_file(stdbool.h HAVE_STDBOOL_H)
 check_include_file(unistd.h HAVE_UNISTD_H)
+check_include_file (immintrin.h HAVE_IMMINTRIN_H)
 
 # For examples and tests
 
--- a/config.h.cmake
+++ b/config.h.cmake
@@ -33,6 +33,9 @@
 /* Define if you have C99's lrintf function. */
 #cmakedefine01 HAVE_LRINTF
 
+/* Define to 1 if you have the <immintrin.h> header file. */
+#cmakedefine01 HAVE_IMMINTRIN_H
+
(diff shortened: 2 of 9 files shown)


erikd (commented):
LGTM.

erikd:
The used to have CI on this project. I think Travis (the old CI) has disappeared, but it should be possible to do something with Github Actions (even if its just Linux, which should be the easiest).

rfomin:
> The used to have CI on this project.

It looks like CI is here: https://github.com/libsndfile/libsamplerate/blob/master/.github/workflows/build.yml
Someone just needs to approve running workflow, I'm first time contributor.

evpobr (approved):

evpobr:
Errors are unrelated to PR. We need to update CI.

evpobr:
Thanks @rfomin !

sezero:
The patch here might be OK for x86_64, but it doesn't give us disabling SSE2
code for x86 builds: I suggest some configury switch like --enable-sse2 (and an
equivalent of it in cmake.)

Issue #191, opened 2023-01-27
fail to build with Xcode in universal binary on OSX


otristan:
libsamplerate failed to build on xcode with arm64 x86_64 architecture

In order to fix the issue, you should add

#if defined(__i386__) || defined(__x86_64__)

where 	#ifdef HAVE_IMMINTRIN_H
is tested
or don't rely on HAVE_IMMINTRIN_H

Thanks !

evpobr:
Hi @otristan , thanks for report.

sezero:
https://github.com/libsndfile/libsamplerate/pull/188 is the guilty party for that.

That patch was made solely because mingw x64 version used x87 asm instead of SSE2,
an as it is, it is causing more problems than it solves - here is a build error on i686-linux:
```
[  1%] Building C object src/CMakeFiles/samplerate.dir/samplerate.c.o
In file included from /home/sezero/libsamplerate/src/samplerate.c:19:
/home/sezero/libsamplerate/src/common.h: In function ‘psf_lrintf’:
/home/sezero/libsamplerate/src/common.h:176: warning: implicit declaration of function ‘_mm_cvtss_si32’
/home/sezero/libsamplerate/src/common.h:176: warning: implicit declaration of function ‘_mm_load_ss’
/home/sezero/libsamplerate/src/common.h: In function ‘psf_lrint’:
/home/sezero/libsamplerate/src/common.h:185: warning: implicit declaration of function ‘_mm_cvtsd_si32’
/home/sezero/libsamplerate/src/common.h:185: warning: implicit declaration of function ‘_mm_load_sd’
[  3%] Building C object src/CMakeFiles/samplerate.dir/src_linear.c.o
In file included from /home/sezero/libsamplerate/src/src_linear.c:19:
/home/sezero/libsamplerate/src/common.h: In function ‘psf_lrintf’:
/home/sezero/libsamplerate/src/common.h:176: warning: implicit declaration of function ‘_mm_cvtss_si32’
/home/sezero/libsamplerate/src/common.h:176: warning: implicit declaration of function ‘_mm_load_ss’
/home/sezero/libsamplerate/src/common.h: In function ‘psf_lrint’:
/home/sezero/libsamplerate/src/common.h:185: warning: implicit declaration of function ‘_mm_cvtsd_si32’
/home/sezero/libsamplerate/src/common.h:185: warning: implicit declaration of function ‘_mm_load_sd’
[...]

PR #193, opened 2023-02-02
use sse2 intrinsics for lrint/lrintf only on windows x64

sezero:
Fixes https://github.com/libsndfile/libsamplerate/issues/187
Fixes https://github.com/libsndfile/libsamplerate/pull/188
Fixes https://github.com/libsndfile/libsamplerate/issues/191

CC: @rfomin

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,7 +88,6 @@ endif()
 
 check_include_file(stdbool.h HAVE_STDBOOL_H)
 check_include_file(unistd.h HAVE_UNISTD_H)
-check_include_file (immintrin.h HAVE_IMMINTRIN_H)
 
 # For examples and tests
 
--- a/config.h.cmake
+++ b/config.h.cmake
@@ -33,9 +33,6 @@
 /* Define if you have C99's lrintf function. */
 #cmakedefine01 HAVE_LRINTF
 
-/* Define to 1 if you have the <immintrin.h> header file. */
-#cmakedefine01 HAVE_IMMINTRIN_H
-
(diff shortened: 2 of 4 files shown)


rfomin:
LGTM

evpobr:
Do we really need to limit SSE2 to x64?

https://github.com/libsndfile/libsndfile/blob/71f7bde180ade1869542d36483bc2e2c809c2189/cmake/CheckCPUArch.cmake#L17-L23

sezero:
> Do we really need to limit SSE2 to x64?

No, we don't. But can you guarantee that `__SSE2__` is defined on x86?

sezero:
> > Do we really need to limit SSE2 to x64?
> 
> No, we don't. But can you guarantee that `__SSE2__` is defined on x86?

E.g.: see the build error I noted in #191

sezero:
Any verdict ?

evpobr:
Not always: https://stackoverflow.com/questions/18563978/detect-the-availability-of-sse-sse2-instruction-set-in-visual-studio

evpobr:
> Any verdict ?

Not sure redefining lrintf is good idea.

sezero:
> Not sure redefining lrintf is good idea.

I don't see the harm since we're doing it privately - but if you want static inlines, tell me.

sezero:
Anyways, I force-pushed a version that keeps the inlines as inlines - no redefines.

evpobr (approved):

evpobr:
Thanks @sezero !

Issue #176, opened 2022-01-13
Parallelize resampling (with high delay)


arlofaria:
Hi @arlofaria.

I like this idea, but we still have no developer to implement this. Maybe you can help us?

_Originally posted by @evpobr in https://github.com/libsndfile/libsamplerate/issues/6#issuecomment-958652191_

arlofaria:
This would require multi-threading, which might be complicated to introduce in this codebase.

ss3git:
I've implemented a multi-thread version for my personal needs.
Sharing it for those who might be interested.

https://github.com/ss3git/libsamplerate/compare/master...ss3git:libsamplerate:MultiThreading

ss3git:
Hi, I hope I sent the request correctly.
I'm actually not so familiar with github yet.

luzpaz:
This looks awesome. What's type of speed boost does this actually give ?

ss3git:
If asking about my implementation, the sinc converter uses multiple threads during its conversion process so is expected to run faster (give more throughput) on multi-core CPUs unless input data blocks to src_process() are very small.

Here is an example comparison of the benchmark (multichan_throughput_test) results on an 8 core cpu.

```
single-thread:
    Converter                        Channels    Duration      Throughput
    ---------------------------------------------------------------------
    Fastest Sinc Interpolator           1          5.03         2914383
    Fastest Sinc Interpolator           2          5.00         2673097
    Fastest Sinc Interpolator           3          5.02         1259082
    Fastest Sinc Interpolator           4          5.01         2072984
    Fastest Sinc Interpolator           5          5.01          974272

    Medium Sinc Interpolator            1          5.06         1255958
    Medium Sinc Interpolator            2          5.03         1186395
    Medium Sinc Interpolator            3          5.03          576003
    Medium Sinc Interpolator            4          5.02          911986
    Medium Sinc Interpolator            5          5.02          433959

    Best Sinc Interpolator              1          5.27          369097
    Best Sinc Interpolator              2          5.13          341291
    Best Sinc Interpolator              3          5.18          175364
    Best Sinc Interpolator              4          5.11          247623
    Best Sinc Interpolator              5          5.17          130465
    Best Sinc Interpolator              6          5.02          215596
    Best Sinc Interpolator              7          5.09           94618
    Best Sinc Interpolator              8          5.12           88757
    Best Sinc Interpolator              9          5.00           83624
    Best Sinc Interpolator             10          5.00           77856

[...]

PR #201, opened 2023-09-20
Multi threading

ss3git:
Experimental multi-thread implementation. (disabled by default with BUILD_MULTI_THREADING build option.)

Related issue: Parallelize resampling (with high delay) #176

compiled and tested on

FreeBSD clang version 14.0.5 (https://github.com/llvm/llvm-project.git llvmorg-14.0.5-0-gc12386ae247c)
Target: x86_64-unknown-freebsd13.2
Thread model: posix

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -72,6 +72,11 @@ if(LIBM_REQUIRED)
     target_link_libraries(samplerate PRIVATE m)
 endif()
 
+if(MULTI_THREADING)
+  target_link_libraries(samplerate PRIVATE omp)
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS_RELEASE} -fopenmp")
+endif()
+
 # Set public header
 set_property(TARGET samplerate PROPERTY PUBLIC_HEADER ${PROJECT_SOURCE_DIR}/include/samplerate.h)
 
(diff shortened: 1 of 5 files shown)


luzpaz:
Is there an ETA on this ?

evpobr:
@SoapGentoo , @arthurt , what do you think? I guess we can add it when it's ready.

src/CMakeLists.txt
```diff
@@ -43,6 +43,16 @@ add_library(samplerate
 # ALIAS to use if libsamplerate is included from other project with add_subdirectory()
 add_library(SampleRate::samplerate ALIAS samplerate)
 
+if(MULTI_THREADING)
+  if(WIN32)
```
evpobr, line 47: I guess we need  to check `MSVC` variable. MinGW compiler probably will fail.

ss3git: commit: 7348582e57dcf5c0ee65ff104f413893648489a8

luzpaz:
Just wondering what's left here ?

ss3git:
Hello @luzpaz.

This PR works well enough at least on my machine.
Maybe waiting for more testers other than me?

evpobr:
@ss3git , are the results bitwise identical to the reference implementation?

ss3git:
Hi, @evpobr .

I've already forgotten a lot of the details though, the core calculation part is intended to be bitwise identical to the reference implementation (if you mean the single-thread version).

Since the single-thread implementation seems to have input/output processing variance depending on the number of channels processed (e.g. https://github.com/libsndfile/libsamplerate/issues/206), I am not hundred percent sure if the outputs "always" match.

evpobr:
> One thing I'm aware of is that if you are actually releasing it, adding converter types like "SRC_SINC_BEST_QUALITY_MT" may be an idea to have both ST and MT versions switchable in one library.

Good idea.

luzpaz:
Sorry to bump, just wondering if there has been any more progress ?

PR #203, opened 2023-11-02
Update build instructions in README.md

alexanderwerning:
The build failed until I added the `autoreconf -i` command (similar to the soundfile build).
I am not familiar with c building/configuring tools, so maybe I missed something.


evpobr (changes requested):
It's better to change to:

```
autoreconf -vif
```

evpobr (approved):

evpobr:
Thanks @alexanderwerning !

Issue #208, opened 2024-03-06
Out of bound array access in src_linear.c


lkoenig:
In some particular cases, `src_linear.c` perform an access on the input array with an index of -1 in https://github.com/libsndfile/libsamplerate/blob/master/src/src_linear.c#L122 when `priv->in_used` is 0 because.

I'll add a PR with a test case catching this corner case.

evpobr:
Hi @lkoenig

lkoenig:
Hi @evpobr :-)

PR #209, opened 2024-03-06
Fix out of boundary access in `src_linear.c` 

lkoenig:
Add a test case with different consecutive buffer size and a fix for #208.

### DIFF
--- a/src/src_linear.c
+++ b/src/src_linear.c
@@ -32,9 +32,7 @@ static void linear_close (SRC_STATE *state) ;
 
 typedef struct
 {	int		linear_magic_marker ;
-	bool	dirty ;
-	long	in_count, in_used ;
-	long	out_count, out_gen ;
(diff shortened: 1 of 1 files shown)


JoergAtGithub:
Is there any reason, why this shouldn't be merged?

lkoenig:
None I can see.

rmpowell77:
+1

rmpowell77 (commented):
Looks great to me.

1820893135-pixel:
Related: #235 is a minimal fix for the same `in_used == 0` case this PR addresses (see #234 for the reproducer). It only adds the `prev = (in_used >= channels) ? data_in[...] : last_value[ch]` fallback to the main loop rather than rewriting the converter, so the two are alternative approaches - whichever the maintainers prefer, the other becomes redundant.

PR #210, opened 2024-03-07
Introduce single precision float compilation option

KKopyscinski:
This patch aims to add possibility for library to be compiled using single precision floating point numbers. This allows to use this library on embedded systems, where FPU may not support double precision floating point math, thus, is resource intensive and to slow for audio processing.


evpobr:
Hi @KKopyscinski

KKopyscinski:
@evpobr Hi :) is this under review?

sjanc (approved):

sjanc:
Hi,

Any chances that this could be merged and included in 0.2.3?
This is crucial for using library on MCUs that don't have HW support for double..

